PR #50, opened 2021-02-24
Implement US metadata checks

sbfnk:
This implements stricter metadata checks as in the US (discussed in #5) whilst preserving previous adaptation to ECDC hub specifics. Currently fails because no metadata is supplied with the forecasts from the German/Polish hub.


sbfnk:
Merging for now - to be updated with further improvements/fixes, see https://github.com/JADEUSC/covid19-forecast-hub-europe/pull/1

PR #52, opened 2021-02-26
Create locations_eu.csv

nikosbosse:


jbracher:
There is already this https://github.com/epiforecasts/covid19-forecast-hub-europe/blob/main/template/locations_eu.csv and [this](https://github.com/epiforecasts/covid19-forecast-hub-europe/blob/main/viz/location_codes.csv). What's the best place to have these info? Warning: The file in viz is read in by the Angular app, so would need to let Fabian know in case we move it.

sbfnk:
There's also this https://github.com/epiforecasts/covid19-forecast-hub-europe/blob/extended-metadata-check/data-locations/locations_eu.csv

The file in `template` is also referenced in the wiki, I think. To be honest I find having a separate `data-locations` folder a bit excessive. Shall we stick with `data-locations` and point Fabian to it (until he's made the change we can keep the locations)?

As a general point for these kind of data files I think we need to have either a reference or code how they were created, e.g. https://github.com/epiforecasts/covid19-forecast-hub-europe/blob/14b02561a6de270efab54abac9314e15f758a079/template/generate_template.R#L20

If it's important to have this now we can merge it, but otherwise we can close and the file will come in when we merge the metadata checks.

nikosbosse:
So I just copied the file directly from the metadata PR, because I thought it would be quicker to merge it separately. But happy either way.

PR #73, opened 2021-03-04
Fix #70: Error if forecast_date <5 days from target_end_date

### LINKED ISSUE #70: Date validation checks
Currently it seems possible to pass a validation check with a forecast date of 2 March (which should correspond to a submission week ending 8 March) and a 1wk ahead target end date of 6 March (instead of 13 March). This should not be possible.

kathsherratt:
Adds extra check to return error and terminate validation if `forecast_date` is too close in time (<5 calendar days) to `target_end_date`. Should solve the problem of invalid submissions with a forecast date Tuesday or later which have a "1 wk ahead" target for that same week's Saturday. Fixes #70

### DIFF
--- a/code/validation/covid19.py
+++ b/code/validation/covid19.py
@@ -100,17 +100,24 @@ def covid19_row_validator(column_index_dict, row, codes):
         except ValueError:
             pass  # ignore, caught by `json_io_dict_from_quantile_csv_file()`
 
-    # 3. validate forecast_date and target_end_date date formats
+    # 3. validate forecast_date and target_end_date
     forecast_date = row[column_index_dict['forecast_date']]
     target_end_date = row[column_index_dict['target_end_date']]
+    
+    # 3.1 Expect date formats
     forecast_date = _parse_date(forecast_date)  # None if invalid format
     target_end_date = _parse_date(target_end_date)  # ""
-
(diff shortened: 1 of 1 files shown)


code/validation/covid19.py
```diff
@@ -100,17 +100,24 @@ def covid19_row_validator(column_index_dict, row, codes):
...
     if not forecast_date or not target_end_date:
         error_messages.append(f"Error > invalid forecast_date or target_end_date format. forecast_date={forecast_date!r}. "
                               f"target_end_date={target_end_date}. row={row}")
-        return error_messages  # terminate - depends on valid dates
-
+        return error_messages
+
+    # 3.2 Expect a minimum 5 days between forecast_date and target_end_date
+    date_diff = target_end_date - forecast_date
+    if date_diff < 5:
```
sbfnk, line 117: This causes
image
I think you might need `date_diff.days` or something like that.

sbfnk (changes requested):
Great stuff, thanks! One minor comment from the PR check failure.

code/validation/covid19.py
kathsherratt: Ah thanks, silly mistake! Fixed.

sbfnk (approved):

PR #82, opened 2021-03-08
2021-03-08 UNIPV-BayesINGARCHX

Barb-00:


sbfnk:
Hi, sorry you're having problems with your checks. I think this is related to your metadata file so should only happen once.

I think there might be a problem with text encoding / special characters
image

Can you try to edit so that new lines show up correctly and there are no special characters (e.g., "Ò", "Ð")? I think you can edit the file directly in the GitHub web interface, e.g. by clicking "Files changed" above and then "edit file". 
image

sbfnk:
Nearly there! You'll need to add a `website_url:` field to the Metadata, and make sure the `license` is one of the allowed values (can be `other`): https://github.com/epiforecasts/covid19-forecast-hub-europe/blob/main/code/validation/accepted-licenses.csv

PR #84, opened 2021-03-08
2021-03-08 UNIPV-BayesINGARCHX

Barb-00:


kathsherratt:
Hi @Barb-00 - thank you for submitting to the hub! 

Before we can get this merged in, it looks like you may need to change the following:
- The license in your metadata file is currently `UNIPV`. However, licenses should be one of [these](https://github.com/epiforecasts/covid19-forecast-hub-europe/blob/main/code/validation/accepted-licenses.csv) (the short license code). We strongly encourage using one of the suggested public licenses to ensure forecasts can be used later on. However if you cannot use one of these, please use "other" in this field.
- You will need to add a `website_url:` key. Ideally this would link to somewhere with a bit more information about your model or team - you could link to a github repo for example.

Again, thanks very much for submitting!
Kath

PR #85, opened 2021-03-08
Adding FIAS and ITWM forecasts (from DE/PL hub)

jbracher:
Also contains minor correction of app to check submissions (sorry for mixing up things).


kathsherratt:
Thanks Johannes!

PR #87, opened 2021-03-08
Simplified version of our model

jlaznarte:
We test the forecasts sending procedure with a simplified version (only 1 week ahead, point forecasts) of our model.


kathsherratt:
Hi @jlaznarte - thanks very much for contributing! 

This looks good, you will just need to change a couple more things to get this passing the checks:
- In the metadata, license needs to be one of the short codes in the [accepted licenses](https://github.com/epiforecasts/covid19-forecast-hub-europe/blob/main/code/validation/accepted-licenses.csv) (this would be "cc-by-sa-4.0")
- It looks like you have multiple values for `forecast_date` ('2021-03-10', '2021-03-08')
   - Fixing this should also fix the other error in the validation ("target_end_date was <5 days after forecast_date")

Then we will be good to go.

jlaznarte:
Thanks, Kath, I think it should be fine now...

kathsherratt:
Nice one, thanks again!

PR #92, opened 2021-03-08
Update metadata-epiforecasts-EpiNow2.txt

seabbs:
Updates EpiNow2 metadata


seabbs:
I feel sad that @nikosbosse is the primary model 😭 when `EpiNow2` has been shown to be between marginally and significantly worse than a NULL model. It just seems unreasonable....

PR #103, opened 2021-03-09
Add a data status page

kathsherratt:
It may become difficult to keep track of problems in data across Europe. I've previously found that most issues get reported to the [JHU issues page](https://github.com/CSSEGISandData/COVID-19/issues), but this gets so many reports on US data that it can take some trawling through to find relevant issues. 

This isn't super necessary but in case it's useful to get them all in one place, this Rmd scrapes the JHU repo for issues that are open, have a European country in the issue title, and opened in the last 8 weeks.

Preview the page [here](http://htmlpreview.github.io/?https://raw.githubusercontent.com/epiforecasts/covid19-forecast-hub-europe/606f06346570e8d879b46f6dcbbb93b276fa8049/data-truth/data-issues.html). It's not very sophisticated (but equally, there may not be that much that can be done) - ideas welcome.

If it looks okay we could potentially just add running this to the daily gh actions.


seabbs:
Very clever - nice!

seabbs (approved):
Just to give it a nice tick.

PR #104, opened 2021-03-12
Add validation check: values are >= 0

kathsherratt:
Checks values are >= 0. 

Currently failing forecasts will always cause all checks to fail unless we do something about them. I think these are the options but none ideal so glad for other ideas:
- remove the entire csv
- remove only the quantiles with negative values
- replace negative values with 0
- keep as is; make the validation check conditional on a `forecast_date` later than 2021-03-08

Files failing on this check, in `data-processed` :
```
itwm-dSEIR/2021-02-22-itwm-dSEIR.csv
itwm-dSEIR/2021-02-15-itwm-dSEIR.csv
itwm-dSEIR/2021-03-01-itwm-dSEIR.csv
itwm-dSEIR/2021-03-08-itwm-dSEIR.csv
epiforecasts-EpiExpert/2021-03-08-epiforecasts-EpiExpert.csv
Karlen-pypm/2021-02-28-Karlen-pypm.csv
Karlen-pypm/2021-03-07-Karlen-pypm.csv
EuroCOVIDhub-ensemble/2021-03-08-EuroCOVIDhub-ensemble.csv
```

### DIFF
--- a/code/validation/covid19.py
+++ b/code/validation/covid19.py
@@ -62,6 +62,7 @@ def covid19_row_validator(column_index_dict, row, codes):
     """
     Does COVID19-specific row validation. Notes:
     - Checks in order:
+    0. value
     1. location
     2. quantiles
     3. forecast_date and target_end_date (terminates if invalid)
@@ -83,7 +84,7 @@ def covid19_row_validator(column_index_dict, row, codes):
     
     # 0. Validate forecast value
     value = row[column_index_dict['value']]
-    if value < 0:
+    if float(value) < 0:
         error_messages.append(f"Error > negative value in forecast: {value!r}. row={row}")
 
     # 1. validate location (ISO-2 code)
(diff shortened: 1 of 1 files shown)


sbfnk:
Good catch! @jbracher any thoughts? Have you encountered this before? My inclination would be to

> keep as is; make the validation check conditional on a forecast_date later than 2021-03-08

so as not to tamper with submitted forecasts but all the suggested options are sensible.

jbracher:
Do I understand correctly that this check was not in place when these files were submitted?

A problem I see is that the truth sometimes *is* negative (e.g. in Spain). Or would we also set it to zero in that case?

I think all suggested solutions are reasonable at this stage. My preference depends on what we decide for truth < 0.

sbfnk:
I don't think there's much we can do about the truth data except changing the data source which is under review (as setting the data to 0 would give the wrong cumulative).

Looking at the (useful!) info in the PR check fail the only ones below 0 beyond rounding are Karlen and itwm - might be getting in touch to let them know, and check if they have a preference for how this is handled.

JADEUSC:
One minor correction from my side: If we include the check in the valdations V2 and start to use them starting from the next submission deadline it would correspond to "make the check conditional on the 2021-03-22".

sbfnk:
Closing as now implemented in https://github.com/JADEUSC/covid19-forecast-hub-europe-validations/issues/1

PR #105, opened 2021-03-12
Zoltar edit

joseph-palmer:
### Overview 

Had a go at getting the processed data on to Zoltor using the [zoltar_scripts](https://github.com/reichlab/covid19-forecast-hub/tree/master/code/zoltar_scripts) from [covid19-forecast-hub](https://github.com/reichlab/covid19-forecast-hub). Setting up with the correct credentials and altering the project name and some file extensions I was able to get the models on to the zoltar page with the metadata: https://zoltardata.com/project/238. However, the forecasts themselves fail to upload. 

### Issues 

The reason for the errors is due to a function in zoltpy ` validate_quantile_csv_file()` ([code](https://github.com/reichlab/zoltpy/blob/313ebb6ff40f7fe507de834c920ef77fbe19afaf/zoltpy/covid19.py)) which checks the contents of a forecast csv file against criteria (column names, values within ranges etc.). All forecasts fail due to an 'invalid location'  and 'invalid quantile' when running `upload_covid19_forecasts_to_zoltar.py`, some also fail for having an extra column, scenario_id, which is not on its list. The validations checking this are written into the zoltpy package. 

### Potential solutions   

1) Ignore the validations and upload the files as they are without checks, can be done I think by running:  `json_io_dict_from_quantile_csv_file()` with the argument `row_validator=None`.   

2) Write a custom `validate_quantile_csv_file()` function to enforce custom checks. This will require custom locations also.   

Happy to have a crack at either but wanted to get thoughts first.


kathsherratt:
Thanks Joe, this is great! We had to customise a lot of the validation functions, so I guess it makes sense that this wouldn't work straight out of the Zoltar box, which is designed to work with the US hub validation process. 

I think this should be fairly straightforward to solve. Where it imports validation functions used by the US hub - i.e. `from zoltpy.covid19 import COVID_TARGETS, COVID_ADDL_REQ_COLS, covid19_row_validator, validate_quantile_csv_file` - we should just need to change this to importing the equivalent functions from our own validation code. 

These are all saved in `code/validation/covid19.py` ([here](https://github.com/epiforecasts/covid19-forecast-hub-europe/tree/main/code/validation/covid19.py)), and what we need to import should have the same or very similar names.  I was mainly responsible for editing these functions though and it may not be clear how the names map between the US vs what we have. So feel free to have a look at this, but it's not urgent and I'm very happy to make this switch over in any case.

Thank you for figuring this part out and fixing it up already!

joseph-palmer:
I changed the function connections to the custom ones and this seems to work. On running the script I get the following for all forcasts:

```
* polling for status change. job: ('Job', 'https://zoltardata.com/api/job/48265/', 48265)
- QUEUED
- FAILED
x FAILED

 _upload_forecast_worker(): error: RuntimeError("The data format of `value` should correspond or be translatable to the `type` as in the target definition, but one of the value values was not. values=[45208.3, 50776.7, 53930.6, 59564.4, 64198.5, 67680.6, 70640.2, 73507.2, 76309.2, 78772.2, 81151.2, 83385, 85615.9, 88045.8, 91063.1, 93621.9, 95857.5, 99326.4, 103767.5, 108981.9, 115956, 122786.9, 131254.3], prediction_dict={'unit': 'DE', 'target': '1 wk ahead inc case', 'class': 'quantile', 'prediction': {'quantile': [0.01, 0.025, 0.05, 0.1, 0.15, 0.2, 0.25, 0.3, 0.35, 0.4, 0.45, 0.5, 0.55, 0.6, 0.65, 0.7, 0.75, 0.8, 0.85, 0.9, 0.95, 0.975, 0.99], 'value': [45208.3, 50776.7, 53930.6, 59564.4, 64198.5, 67680.6, 70640.2, 73507.2, 76309.2, 78772.2, 81151.2, 83385, 85615.9, 88045.8, 91063.1, 93621.9, 95857.5, 99326.4, 103767.5, 108981.9, 115956, 122786.9, 131254.3]}}",)
RuntimeError occured while uploading forecast. Error: job failed: job=('Job', 'https://zoltardata.com/api/job/48265/', 48265, 'FAILED'), failure_message='_upload_forecast_worker(): error: RuntimeError("The data format of `value` should correspond or be translatable to the `type` as in the target definition, but one of the value values was not. values=[45208.3, 50776.7, 53930.6, 59564.4, 64198.5, 67680.6, 70640.2, 73507.2, 76309.2, 78772.2, 81151.2, 83385, 85615.9, 88045.8, 91063.1, 93621.9, 95857.5, 99326.4, 103767.5, 108981.9, 115956, 122786.9, 131254.3], prediction_dict={\'unit\': \'DE\', \'target\': \'1 wk ahead inc case\', \'class\': \'quantile\', \'prediction\': {\'quantile\': [0.01, 0.025, 0.05, 0.1, 0.15, 0.2, 0.25, 0.3, 0.35, 0.4, 0.45, 0.5, 0.55, 0.6, 0.65, 0.7, 0.75, 0.8, 0.85, 0.9, 0.95, 0.975, 0.99], \'value\': [45208.3, 50776.7, 53930.6, 59564.4, 64198.5, 67680.6, 70640.2, 73507.2, 76309.2, 78772.2, 81151.2, 83385, 85615.9, 88045.8, 91063.1, 93621.9, 95857.5, 99326.4, 103767.5, 108981.9, 115956, 122786.9, 131254.3]}}",)'
```

After this I ran the code a second time to debug but the message had gone, not sure why it arrises in the first place but it seams it uploads the forcasts regardless. Only some of the the forcasts were up when I checked, but this also happened when I uploaded the models, some came on immediatly, others took longer.

sbfnk:
Are we concluding that this is working now? Do we want a github action for regular uploads?

sbfnk:
Just bumping this so it doesn't get forgotten.

kathsherratt:
This now works when running locally - but needs testing with github action #570.

sbfnk:
Great stuff! I'm lost on the detail but I guess the thing to do will be to run it in parallel for a bit and then check that everything is there, files are the same etc.

sbfnk (approved):

code/zoltar_scripts/README.md
```diff
@@ -0,0 +1,36 @@
...
+#### Local upload to Zoltar
+
+First ensure you have set up  hub Zoltar access by adding `Z_USERNAME` and `Z_PASSWORD` as environment variables.
+
+```
+# Clone repo with validations submodule
+git clone epiforecasts/covid19-forecast-hub-europe --recurse-submodules
+
+# Set hub as working directory, e.g.
+# cd Github/covid19-forecast-hub-europe
```
sbfnk, line 19: minimally minor comment but would just make it `cd covid19-forecast-hub-europe` as everything else depends on local structure.

code/zoltar_scripts/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+# Detect modified files, convert to json and upload
+python code/zoltar_scripts/upload_zoltar.py
+```
+Detection, validation, and upload takes ~10 seconds per forecast.
```
sbfnk, line 24: per forecast file?

code/zoltar_scripts/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+#### Other tasks
+
+ - Compare forecasts in the repo with those in Zoltar with:
+```
+ python code/zoltar_scripts/compare_repo_to_zoltar.py
+ ```
+
+- Create json file with:
```
sbfnk, line 33: What's the purpose of this?

code/zoltar_scripts/compare_repo_to_zoltar.py
```diff
@@ -0,0 +1,44 @@
...
+import os
+
+# 1 Meta info 
+
+## Set connections to Zoltar project and github repo
+project_name = 'ECDC European COVID-19 Forecast Hub'
+project_obj = None
+project_timezeros = []
+conn = util.authenticate()
+url = 'https://github.com/epiforecasts/covid19-forecast-hub-europe/tree/master/data-processed/'
```
sbfnk, line 11: It might be good to keep a list somewhere of things that would be good to store in a [config file](https://github.com/epiforecasts/covid19-forecast-hub-europe/pull/511) but aren't a priority to implement.

code/zoltar_scripts/compare_repo_to_zoltar.py
```diff
@@ -0,0 +1,44 @@
...
+print("number of forecasts in repo: " + str(len(repo_forecasts)))
+for forecast in zoltar_forecasts:
+    if forecast not in repo_forecasts:
+        print("This forecast in zoltar but not in repo "+forecast)
+print()
+print('----------------------------------------------------------')
+print()
+for forecast in repo_forecasts:
+    if forecast not in zoltar_forecasts:
+        print("This forecast in repo but not in zoltar "+forecast)
```
sbfnk, line 44: This only compares the file names, right? It might be good to do some spot checks (doesn't have to be implemented here) to check that the files are actually the same.

code/zoltar_scripts/upload_zoltar.py
```diff
@@ -0,0 +1,161 @@
+from __future__ import absolute_import
+
+# from zoltpy.quantile_io import json_io_dict_from_quantile_csv_file
+from zoltpy import util
+from zoltpy.connection import ZoltarConnection
+# from zoltpy.covid19 import COVID_TARGETS, COVID_ADDL_REQ_COLS, covid19_row_validator, validate_quantile_csv_file
```
sbfnk, line 6: remove?

code/zoltar_scripts/upload_zoltar.py
```diff
@@ -0,0 +1,161 @@
+from __future__ import absolute_import
+
+# from zoltpy.quantile_io import json_io_dict_from_quantile_csv_file
```
sbfnk, line 3: remove?

code/zoltar_scripts/upload_zoltar.py
```diff
@@ -0,0 +1,161 @@
...
+cwd_p = Path(__file__).parent.resolve()
+all_forecasts = glob.glob('./data-processed/**/*-*.csv')
+pprint.pprint(all_forecasts)
+# meta info
+project_name = 'ECDC European COVID-19 Forecast Hub'
+project_obj = None
+conn = util.authenticate()
+url = 'https://github.com/epiforecasts/covid19-forecast-hub-europe/tree/main/data-processed'
+
+# all_forecasts = glob.glob('./data-processed')
```
sbfnk, line 35: remove?

sbfnk (approved):
In addition to the few comments: Can we add a brief explanation somewhere (e.g. a README in the directory) of what `validated_file_db.json` and `validated_file_db.p` are?

code/zoltar_scripts/README.md
kathsherratt: It's unlikely to ever be used (hopefully), but if for some reason the python-zoltar connection is not working, then this script creates the correctly-formatted json file which can be uploaded manually using the zoltar web interface. I've added this to the README.

code/zoltar_scripts/upload_zoltar.py
kathsherratt: yes, sorry, these are leftovers from adapting from US hub / Joe's code, thanks for flagging.

Bisaloo (commented):
I see a file called `validated_file_db.p` but as far as I can tell, it's not used anywhere. Is this the result of a typo?

kathsherratt:
> I see a file called `validated_file_db.p` but as far as I can tell, it's not used anywhere. Is this the result of a typo?

The `.p` extension is a python pickle file which goes with the `json`. From my understanding, these files are tracking the changes to add to Zoltar so (I think) need to be kept, though I'll check.

code/zoltar_scripts/compare_repo_to_zoltar.py
Bisaloo: @kathsherratt, you can now use the `repo_url` for the config file `forecasthub.yml` if you wish.

Bisaloo (approved):

kathsherratt:
Unfortunately I couldn't get this to work when run automatically from github actions, because of some difficulty with python importing from modules in the validation module. 

I have suggested a solution which runs both locally and, as far as I can tell, runs with git action as well: this is by moving the zoltar code directly to the validations repo [here](https://github.com/epiforecasts/covid19-forecast-hub-europe-validations/pull/18) (and updating the github action to reflect the new filepath [here](https://github.com/epiforecasts/covid19-forecast-hub-europe/pull/639). 

This means the git action runs using code inside the validation submodule only, as this way we can access both the validation code and the repo files in data-processed without changing the origin of relative file paths.

I'm sorry this has been such a convoluted process. I think the two PRs linked above should, finally, do the trick, and we can close this PR completely.

PR #118, opened 2021-03-15
Forecast update 2021-03-15 | Priesemann

semohr:


semohr:
Seems like the github actions service is down (see [here](https://www.githubstatus.com/incidents/s654n76c1bwr)). Feel free to rerun the check once it is working again. Maybe a [self-hosted](https://docs.github.com/en/actions/hosting-your-own-runners) runner could circumvent this in the future, never saw an action outage before though.

PR #145, opened 2021-03-22
2021-03-22 UPG.UMIB.USI-MULTI.BAYES

miraanto:
First submission UPG.UMIB.USI-MULTI.BAYES


sbfnk:
Hi, thanks for your submission! 

I think you have a colon(`:`) in your metadata file that confuses the validator. In line 8 of `metadata-UPG.UMIB.USI-MULTI.BAYES.txt`, after `categories`.

sbfnk:
I think there are two more problems:
1. Can you replace the dots (`.`) in the team and model name with underscores (`_`)? I think you have to do this both in the directory name and in the metadata (`model_abbr`)
2. The `methods` field is too long: max. 200 characters.

miraanto:
Thanks for your help
very useful!
I think we have successfully completed the procedure - can you pls confirm?
One question that still remains is the following: we submit today the forecast for the number of cases for the week that starts today or that started yesterday?
Thanks
Antonietta

PR #157, opened 2021-03-25
Ensemble code refresh + median ensemble

kathsherratt:
This aims to refresh the ensemble code to make it easier to see what's happening and add new ensembling methods in future. It also adds a new hub ensemble using the median.

The main motivation for this is wanting to split out the inclusion criteria to make it clearer, and if we want to edit this for some other reason later on (e.g. only including primary model designations or team has x past submissions). This may not yet be the most elegant way to do this, but hopefully a start. So far I've included four criteria from the existing code (in `simple-averages.R`).

The PR:

1. Cleans up ensembling code
    - removed unused code/csvs from copy over from German hub
    - removed currently used `simple-averages.R` 
      - code remixed into `create-ensemble-average.R` and  `format-ensemble.R`

2. Re-organises for modular ensemble building
    - A model can be excluded by adding it to `manual_exclusions.csv` with a forecast date and model name
    - New utility functions:
      - `use_ensemble_criteria()` filters forecasts based on four criteria:
        - not manually specified for exclusion
        - includes all quantiles
        - includes 4 horizons
        - not a Euro hub ensemble model
      - `format_ensemble()` formats an ensemble as a standard submission file (there are some differences after loading in from covidHubUtils)
    - Remixed ensembling functions:
      - `create_ensemble_average()` - mean and median ensembles using previous code from `simple-averages.R` 
   - `run-ensembles.R` then runs and saves ensembles:
      - loads forecasts using covidHubUtils; filters with `manual_exclusions.csv` and `use_ensemble_criteria()`
      - saves the criteria used to include/exclude all models as a dated csv in `weekly-criteria`
      - runs and formats mean and median ensembles with `create_ensemble_average()`; saves ensembles in relevant EuroCOVIDhub folders of `data-processing`
   - Also:
     - Re-wrote folder README for usability
     - Had a go at editing the ensemble github action - the only manual input would then be editing `manual_exclusions.csv` as needed

4. Adds median ensemble 
   - Code part of `create_ensemble_average()` (run in `run_ensembles.R`)
   - Created new `data-processed/EuroCOVIDhub-median` folder
     - Added meta-data (model designation: "other")
     - Added median ensemble from submissions as of `forecast_date` 2021-03-22, excluding a late submission
   - New code was not run on mean ensemble for this week, so no change to latest ensemble (EuroCOVIDhub-ensemble)
   
 
Very open to suggestions!


kathsherratt:
Looks like using the median creates non-monotonic-increasing quantiles - because of the extreme range of the values at the edges of uncertainty.

sbfnk:
This looks great! I think the problem with quantile crossing might be that there's at least one model that was missing some quantiles for some horizons in some of the countries - so I think the inclusion criteria need to be assessed not just by date but by date, location and target. If I read the code correctly, at the moment a model gets included as long as all quantiles are present for some combination of target / location.

code/ensemble/README.md
```diff
@@ -1,7 +1,35 @@
...
+### Ensemble methods
+
+**Current methods**
+
+Ensembling methods combine forecast values by target, location, horizon, and quantile. At the moment we use the following:
+
+- Mean average ([`EuroCOVIDhub-ensemble`](./data-processed/EuroCOVIDhub-ensemble))
+- Median average ([`EuroCOVIDhub-median`](./data-processed/EuroCOVIDhub-median))
+
+The ensemble we use in evaluation is the "EuroCOVIDhub-ensemble".
```
sbfnk, line 24: I'd add something like: "We are continually reviewing the performance of the default ensemble compared to other ensembling methods and will make a change if we find a different method to consistently outperform the current default" or something to that effect.

code/ensemble/utils/use-ensemble-criteria.R
```diff
@@ -0,0 +1,60 @@
...
+                                  forecast_date = floor_date(today(), "week", 1)) {
+  
+  # 1. Identify models with all quantiles
+  quantiles <- round(c(0.01, 0.025, seq(0.05, 0.95, by = 0.05), 0.975, 0.99), 3)
+  
+  all_quantiles <- forecasts %>%
+    group_by(model, target_variable, location, target_end_date) %>%
+    summarise(n_quantiles = n(), 
+              .groups = "drop") %>%
+    filter(n_quantiles != length(quantile)) %>%
```
sbfnk, line 33: I think there's a nifty bug here. `length(quantile)` will take the `quantile` function and therefore always be length 1. Did you mean to write `n_quantiles == length(quantiles)`?

Also it's probably more robust to summarise `all_quantiles_present = setequal(quantile, quantiles)` (as it checks that the quantiles are actually the same, not just same length).

code/ensemble/utils/use-ensemble-criteria.R
```diff
@@ -0,0 +1,60 @@
...
+# 2. 4 horizons 
+# 3. Not manually specified for exclusion 
+# 4. Not the hub ensemble
+# 
+# Returns a list:
+# - "forecasts" = tibble, all forecasts passing inclusion criteria
+# - "criteria" = tibble, model names and criteria assessment
+# 
+library(dplyr)
+library(tibble)
```
sbfnk, line 15: needs library(here)?

code/ensemble/utils/use-ensemble-criteria.R
```diff
@@ -0,0 +1,60 @@
...
+
+use_ensemble_criteria <- function(forecasts = 
+                                    covidHubUtils::load_forecasts(source = "local_hub_repo",
+                                                                  hub_repo_path = here(),
+                                                                  hub = "ECDC",
+                                                                  forecast_dates = forecast_date),
```
sbfnk, line 21: could make forecast_date the first argument to the function here I think? (assuming the second argument has access to the first argument)

code/ensemble/utils/use-ensemble-criteria.R
```diff
@@ -0,0 +1,60 @@
...
+    summarise(n_quantiles = n(), 
+              .groups = "drop") %>%
+    filter(n_quantiles != length(quantile)) %>%
+    pull(model) %>%
+    unique()
+  
+  # 2. Identify models with 4 week forecasts
+  all_horizons <- forecasts %>%
+    group_by(model, target_variable, location) %>%
+    filter(any(grepl("4", horizon))) %>%
```
sbfnk, line 40: This probably works but could make it a bit more robust/consistent by defining `horizons = 1:4` and then check with `setequal` as above.

code/ensemble/utils/use-ensemble-criteria.R
```diff
@@ -0,0 +1,60 @@
...
+  
+  # 1. Identify models with all quantiles
+  quantiles <- round(c(0.01, 0.025, seq(0.05, 0.95, by = 0.05), 0.975, 0.99), 3)
+  
+  all_quantiles <- forecasts %>%
+    group_by(model, target_variable, location, target_end_date) %>%
+    summarise(n_quantiles = n(), 
+              .groups = "drop") %>%
+    filter(n_quantiles != length(quantile)) %>%
+    pull(model) %>%
```
sbfnk, line 34: This still includes any model that passes the check for any of the target variables / locations / target_end_date. I think you'll need to keep the full information here (whether all_quantiles present, i.e. don't do the pull but return the data frame information by target and location).

The edge case we've had is where one model, at one date, for one target is missing some quantiles for one horizon. This model should be excluded from the ensemble for that target/location combination but included for all the other ones.

I think this could be done here by doing (unchecked so might be wrong):
```{r}
all_quantiles <- forecasts %>%
    group_by(model, target_variable, location, target_end_date) %>%
    summarise(all_quantiles_present = setequal(quantile, quantiles)) %>%
    group_by(model, target_variable, location) %>%
    summarise(all_quantiles_present_for_all_horizons = all(all_quantiles_present)))
```

code/ensemble/utils/use-ensemble-criteria.R
```diff
@@ -0,0 +1,60 @@
...
+    unique()
+  
+  # 2. Identify models with 4 week forecasts
+  all_horizons <- forecasts %>%
+    group_by(model, target_variable, location) %>%
+    filter(any(grepl("4", horizon))) %>%
+    pull(model) %>%
+    unique()
+  
+  criteria <- tibble("forecast_date" = forecast_date,
```
sbfnk, line 44: If you keep the data frame and don't pull the model (as I suggest above) I think you can then just do 

```{r}
criteria = all_quantiles %>%
  inner_join(all_horizons, by = c("model", "target_variable", "location") %>%
  [your code from line 49 onwards]
```

sbfnk (changes requested):
I think it's still going wrong in the ensemble criteria - made a few suggestions for changes.

code/ensemble/utils/use-ensemble-criteria.R
kathsherratt: I'm not sure it does have access, so have defined in both places to be safe.

code/ensemble/utils/use-ensemble-criteria.R
kathsherratt: Thanks for spotting! Switched to using `setequal` after dropping point forecasts.

kathsherratt:
Thank you for the suggestions @sbfnk and sorry I have been slow to fix this. Updated the criteria code in line with your ideas (much cleaner and works better). The median forecast produced using this now passes our checks. I think this should be ready to use.

code/ensemble/manual_exclusions.csv
```diff
@@ -0,0 +1,2 @@
+﻿forecast_date,model
```
sbfnk, line 1: Can I suggest we add a column `reason_for_exclusion`?

code/ensemble/manual_exclusions.csv
```diff
@@ -0,0 +1,2 @@
...
+22/03/2021,MIMUW-StochSEIR
```
sbfnk, line 2: Suggest to make the dates standard R format, i.e. `2021-03-22`

code/ensemble/manual_exclusions.csv
sbfnk, line 1: I've had a look through the tagged ensemble commits. I think we also want 
`2021-03-08,ITWW-county_repro,"late submission"`

sbfnk (commented):
Thanks, looking good. I made a few more suggestions. 

I've also been thinking a bit about workflow - I think it would be good to have just one model `EuroCOVIDhub-ensemble` in the visualisations/evaluations rather than multiple ensemble versions. So my preference would be something like:
1. Have a file (e.g. a simple text file or a json or something) somewhere that defines the current ensemble (i.e. currently it could just say "mean")
2. Turn `run-ensembles.R` into a function that takes a date (`forecast_date`, the Monday of a submission week so it would consider the preceding week) and a type of ensemble (e.g., "mean") and returns the correctly formatted forecast. This could then be used both to generate the "current" ensemble (defined in (1)) and for our own purposes to generate any other ensembles (e.g., median or weighted or QRA etc) that we might want to consider/compare and that could then be generated for all past dates. We could even maybe have  a script that runs all ensembles on all past dates and then generates an evaluation report that we could use to compare performance.
3. Have a script that is run through a github action that, every Tuesday, checks what the current ensemble is, produces just that, saves it to the "EuroCOVIDhub-ensemble" folder, and maybe adds a line to a csv file somewhere tracking the current/historical choice of ensemble.

Does that sound sensible?

kathsherratt:
Thanks Seb, that sounds good. It sounds fairly near to what we have set up running in the Epiforecasts US forecasts which is usually quite smooth in the ensembling part at least. 

I'll implement, along with your other comments. On step 2, I will leave out for now the last point re. running on all past forecasts and setting up an evaluation report, which seems like it might be best in a separate PR.

sbfnk:
Great, sounds good to me re leaving step 2 for later.

code/ensemble/run-ensembles.R
```diff
@@ -0,0 +1,81 @@
...
+library(here)
+library(vroom)
+library(lubridate)
+library(covidHubUtils)
+source(here("code", "ensemble", "utils", "use-ensemble-criteria.R"))
+source(here("code", "ensemble", "utils", "format-ensemble.R"))
+
+# Set up ----------------------------------------------------------------
+team_name <- "EuroCOVIDhub"
+forecast_date <- floor_date(today(), "week", 1)
```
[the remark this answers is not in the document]
nikosbosse: there is also now a function in covidhubtuils that just converts the date to the appropriate Monday! I'm using that for the reports and I think it works fine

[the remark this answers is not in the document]
sbfnk: ah great - can you link to an example of where that's happening?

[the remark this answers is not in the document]
nikosbosse: Ah you just recoded it: https://github.com/epiforecasts/covid19-forecast-hub-europe/blob/2dc9d4f4521eadce091c3719e349f2654344de20/reports/evaluation/evaluation-by-country.Rmd#L55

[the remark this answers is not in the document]
nikosbosse: There was an older version that used this function: https://github.com/reichlab/covidHubUtils/blob/06eef72664c10b6b10fef38a9af57f5e60d41674/R/date_management.R#L125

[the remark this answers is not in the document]
sbfnk: Ah cool (though not sure why that function would return a character not a date).

Also worth noting that this solves the inverse problem to what we're trying to do here: find all forecast dates that correspond to a given submission date.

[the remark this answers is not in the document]
nikosbosse: I coded it that way :eyes: but we can change it if we want :)

jbracher:
If I understand the above right there will be a "blacklist" of models to be excluded at a given date. Will there be any easily accessible trace of which models were actually used in a given week? I'm wondering if every now and then there will be a model added to the repo after the ensemble build. Then we sort of lose the information of which models were included.
I'd suggest to just have a text file or similar written out with that info (we have these files: https://github.com/KITmetricslab/covid19-forecast-hub-de/tree/master/code/ensemble/included_models; but that format won't work with 29 countries)

jbracher:
Another question: Is a model either included or excluded for all countries? I.e., if a model does something weird for Iceland we'll also exclude it for Spain?

sbfnk:
> If I understand the above right there will be a "blacklist" of models to be excluded at a given date. Will there be any easily accessible trace of which models were actually used in a given week? I'm wondering if every now and then there will be a model added to the repo after the ensemble build. Then we sort of lose the information of which models were included.

Yes, we'll keep track of exclusion criteria and reasons for exclusion any week in a file like [this one](https://github.com/epiforecasts/covid19-forecast-hub-europe/blob/median-ensemble/code/ensemble/weekly-criteria/2021-03-22.csv) (except it'll be by country). Also the ensemble commits are tagged so it'll always be possible to re-create the exact situation at the time the ensemble was built.

sbfnk:
> Another question: Is a model either included or excluded for all countries? I.e., if a model does something weird for Iceland we'll also exclude it for Spain?

No, if e.g. not all quantiles are present in Iceland but they are in Spain it'll be excluded for Iceland but not for Spain.

jbracher:
Ok, that sounds like a good system. Maybe at a later stage one could even add a column for the ensemble weight received.There's no distinction between cases and deaths though?

sbfnk:
Yes, it'll be separated by target variable, too.

Including a model weight is a good idea.

sbfnk:
Although of course if we generate a median ensemble there is no such thing as a model weight...

kathsherratt:
I've updated in line with review so this is ready for another look.

- Reshaped to a `run_ensemble()` function which can be used with any set of forecast dates/models/methods
- Separated out the hub-specific weekly ensemble creation from the generic ensemble code
   - Added `current-model.txt` as a single-line method; added `model-by-date.csv`; refiled model criteria
   - Weekly run of ensemble and model documentation should now be fully automated in `create-weekly-ensemble.R`, called in the github action for running the ensemble.

Also updated the folder README as a better guide linking to both hub-specific forecasts/criteria/methods as well as the generic ensemble code.

sbfnk (approved):

sbfnk:
All looking good - made some minor changes / fixes and merging now. Great work!

PR #158, opened 2021-03-26
update github visualisation action

sbfnk:
Serves location file from data-locations directory and prepares metadata


dwolffram (commented):
We should probably run prepare_metadata.py before prepare_forecast_data.py because the models designated as "other" in the metadata are excluded from the forecast data.

sbfnk:
Thanks, makes sense

PR #201, opened 2021-04-06
bisop seirfilter + seirfilterlite

cyberklezmer:


kathsherratt:
Hi @cyberklezmer , thanks for submitting and contributing a new model this week!

Please could you give us a little more information on the new `seirfilterlite` model, and the differences between this and the `seirfilter` model? 

In your PR metadata, it looks like both this and the `seirfilter` model would be designated "primary". 

Our policy is to have only one model per team designated as "primary". The other model will then be either "secondary" (methods are clearly different from the "primary" model), or "other" (methods are a variation on the "primary" model).

There is a bit more detail on this in the [Wiki](https://github.com/epiforecasts/covid19-forecast-hub-europe/wiki/Metadata).

PR #202, opened 2021-04-06
Adding files submitted via DE/PL platform.

jbracher:


kathsherratt:
Thanks @jbracher ! 

Looks like a conflict caused by merging the submission that came through directly for `MIMUW-StochSEIR`. Hopefully solved quickly by removing that file from this PR, or merging main into this PR + keeping the file on main.

jbracher:
Hi Kath, thanks for identifying the problem. I was not aware MIMUW would submit directly, will remove their file and try again.

jbracher:
Should be fine now :)

PR #253, opened 2021-04-13
Adding forecasts submitted via DE/PL platform.

jbracher:


kathsherratt:
Hi @jbracher - thanks again for adding these! I think the merge conflict is the same problem as last week with an overlapping submission as the MIMUW team submit themselves. They are doing this every week, so not sure if the transfer across from DE/PL hub is automated but just need to remove them here/in future to fix the merge conflict.

jbracher:
Damn :/ I'll exclude them manually in the script which prepares the files.

jbracher:
Ok, looks like this worked :)

PR #284, opened 2021-04-21
Matadata for seirfilterlite

cyberklezmer:


sbfnk:
Thanks - I think you need to remove the colon (`:`) in line 11 (after "estimation from past") as it confuses the YAML parser.

sbfnk:
Thanks!

PR #298, opened 2021-04-26
updates ICM model metadata

eatrunner:


sbfnk:
Hi - thanks for updating the metadata! I think the problem is that in line 10 you have colons (`:`) which confuses the YAML parser - if you replace them with something else, or use quotes (`"`) around the entry, it should work.

eatrunner:
Thank you Sebastian for help.

PR #332, opened 2021-05-03
2021-05-03 epiMOX-SUIHTER

giovanniardenghi:


sbfnk:
Hi, thanks for your submission! I think there is a problem in your metadata file because there is a colon (`:`) which confuses the YAML parser. To solve this, I think you can just put the whole entry in `methods_long` in quotation marks:
`methods_long: "The results are obtained ... in the model forecast."`

PR #342, opened 2021-05-06
Remove faulty forecast for Slovenia

nikosbosse:
For an unknown reason a faulty forecast was submitted for Slovenia. This PR removes that forecast


sbfnk:
Thanks. Can you also remove the point forecast and the one from EpiExpert_direct?

PR #430, opened 2021-05-25
bisop seirfileter - last errorneous prediction deleted

cyberklezmer:


kathsherratt:
Thanks @cyberklezmer - this does not add or change any files so I will close the PR.

PR #464, opened 2021-06-01
Data submission

semohr:
We renamed our Team. From `Priesemann` to `DSMPG`, I hope this does not create any problems for you.
Best, Sebastian


sbfnk:
Thanks! Shouldn't be a problem.

PR #466, opened 2021-06-01
Exclude data anomalies from evaluation

kathsherratt:
This PR adds a csv noting which locations/target variables have anomalies affecting the weekly evaluation.

These values are excluded from evaluation.

This is really a placeholder before implementing an automated anomaly detection process.

### DIFF
--- a/code/evaluation/score_models.r
+++ b/code/evaluation/score_models.r
@@ -112,10 +112,19 @@ forecasts <- forecasts[forecast_date >= "2021-03-08"]
 setnames(forecasts, old = c("value"), new = c("prediction"))
 
 ## load truth data -------------------------------------------------------------
-truth <- map_dfr(.x = c("inc case", "inc death"),
+raw_truth <- map_dfr(.x = c("inc case", "inc death"),
                  .f = ~ load_truth(truth_source = "JHU",
                                    target_variable = .x,
                                    hub = "ECDC"))
+# get anomalies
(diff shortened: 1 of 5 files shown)


code/evaluation/score_models.r
```diff
@@ -112,10 +112,19 @@ forecasts <- forecasts[forecast_date >= "2021-03-08"]
...
                                    target_variable = .x,
                                    hub = "ECDC"))
+# get anomalies
+anomalies <- read_csv(here("data-truth", "anomalies", "anomalies.csv"))
+truth <- left_join(raw_truth, anomalies,
+                            by = c("target_variable",
+                                   "location", "location_name",
+                                   "target_end_date")) %>%
+  filter(is.na(anomaly)) %>%
+  mutate(anomaly = NULL)
```
sbfnk, line 126: I think I would have done `select(-anomaly)` but if that works it's fine.

sbfnk (changes requested):
Great, thanks! Can you do the same in:
https://github.com/epiforecasts/covid19-forecast-hub-europe/blob/main/code/reports/compile-evaluation-reports.r
https://github.com/epiforecasts/covid19-forecast-hub-europe/blob/main/code/reports/ensemble/ensemble-report.Rmd
(and yes, at some point we should remove these code duplications but it's not urgent)

sbfnk (approved):

PR #472, opened 2021-06-01
Try to speed up package installation in R-based GH actions

Bisaloo:
This is a first attempt at speeding up package installation in R-based GH actions. I might have to revisit it because:

- I've had some troubles getting RStudio Package Manager to actually deliver binaries on Linux in the past
- The achieved speed up might not be sufficient. In that case, I can add a caching step to avoid re-downloading / re-installing these packages each time

### DIFF
--- a/.github/workflows/ECDC.yml
+++ b/.github/workflows/ECDC.yml
@@ -8,6 +8,8 @@ jobs:
   get_ecdc:
     runs-on: ubuntu-20.04
     if: github.repository == 'epiforecasts/covid19-forecast-hub-europe'
+    env:
+      RSPM: "https://packagemanager.rstudio.com/cran/__linux__/focal/latest"
     steps:
     - uses: actions/checkout@master
     - uses: r-lib/actions/setup-r@v1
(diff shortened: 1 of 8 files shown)


sbfnk:
Cool! Is the best way to test it is to merge and give it a go? The forecasts come in Saturday-Tuesday so between now and the end of the week is a good time to try potentially breaking changes.

Bisaloo:
Yes, I think that'd be the best strategy. For the other changes, I can try to add some tests to limit the risk of breakage.

sbfnk:
I'm getting a merge conflict - can you bring this up-to-date with `main`?

Bisaloo:
I'm not sure what happened but looks like you already brought these changes to `main`. I guess we can just close this PR.

sbfnk:
Shaved more than 10 minutes off the runtime of this (which failed for other reasons as there was nothing new to commit): https://github.com/epiforecasts/covid19-forecast-hub-europe/actions/workflows/ECDC.yml

Amazing!

PR #475, opened 2021-06-03
Multi ensemble report

sbfnk:
Not suggesting to merge this as it's more copied & pasted code, just flagging to @kathsherratt that this is what I've been using to score the different ensembles.


sbfnk:
Very similar to https://github.com/epiforecasts/covid19-forecast-hub-europe/blob/main/code/reports/compile-evaluation-reports.r just pulling forecasts from a different directory.

PR #477, opened 2021-06-04
Remove dependency to purrr and misc fixes

Bisaloo:
These are a couple of fixes / simplifications I noticed while trying to get more familiar with the codebase. Let's hope I didn't already break something :crossed_fingers: 

I recommend reviewing this PR commit by commit as I grouped together related changes.

### DIFF
--- a/.github/workflows/evaluation.yml
+++ b/.github/workflows/evaluation.yml
@@ -19,7 +19,7 @@ jobs:
 
     - name: Install R dependencies
       run: |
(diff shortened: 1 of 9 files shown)


kathsherratt:
Ah this is definitely neater, thanks Hugo!

sbfnk:
Thanks, this all looks very sensible. I'll merge after the reports have been generated on Tuesday.

PR #480, opened 2021-06-07
Remove large files from repo

Bisaloo:
A couple of questions / suggestions for further general cleaning. Please let me know if there are any constraints I'm unaware of that make this impossible:

- [ ] `viz/necessary_changes` seems like outdated internal communication and could be deleted
- [ ] `docs/decisions.md` could be moved to the wiki
- [ ] All remaining csv (and json?) files should be stored via git-lfs (since we are not interested in the diffs)
- [ ] Move files from `github-actions/` to `.github/workflows/`?
- [ ] Similarly, move `schema.yml` to `.github/workflows/`?

Things that are impossible at the moment but should be considered in the future:

- [ ] It seems inefficient to store `data-truth` in the repo when it comes almost directly from the ECDC or JHU with minimal changes. But we can't change this at the moment because it is required by `covidHubUtils`
- [ ] `data-locations/locations_eu.csv` is redundant with `covidHubUtils::hub_locations_ecdc()` but this might change in the future if the `hub` argument is dropped from `covidHubUtils`


sbfnk:
Thanks, very useful.

> * [ ]  `viz/necessary_changes` seems like outdated internal communication and could be deleted

agree

> * [ ]  `docs/decisions.md` could be moved to the wiki

I'm OK with moving it elsewhere but I'm not 100% sure that the wiki is the right place. The idea for this was to keep track of things we had to change from the US set up in order to get set up ourselves, so that anyone who wants to mirror the approach in the future knows what decisions/changes they need to make.

> * [ ]  All remaining csv (and json?) files should be stored via git-lfs (since we are not interested in the diffs)

I know next to nothing about this. Very much liking the idea of shrinking the repo but can this be done without disrupting the submission workflow for participants (how would usually update submissions in their fork and then PR but never merge the original repo back in). Also in the csv files we might be interested in the diffs (see my comment about `data-truth` below).

> * [ ]  Move files from `github-actions/` to `.github/workflows/`?

Sounds good to me, or maybe a separate directory somewhere around there so as not to mix workflow files with supporting files in the same dir?

> * [ ]  Similarly, move `schema.yml` to `.github/workflows/`?

Would github not then try to interpret it as instructions for an Action? Also see previous comment.

> Things that are impossible at the moment but should be considered in the future:
 
> * [ ]  It seems inefficient to store `data-truth` in the repo when it comes almost directly from the ECDC or JHU with minimal changes. But we can't change this at the moment because it is required by `covidHubUtils`

Yes, also we might consider/switch to other data sources in the future, and we do want to keep track of past truth data in case they change (i.e., we want to know what the data looked like on day X, even if corrections of past data were made later).

> * [ ]  `data-locations/locations_eu.csv` is redundant with `covidHubUtils::hub_locations_ecdc()` but this might change in the future if the `hub` argument is dropped from `covidHubUtils`

Whatever the solution is I agree that it would be good to be consistent here and avoid duplication (bearing in mind the need of someone who would like to set up a forecast hub for disease X / location Y).

sbfnk (approved):
The proposed changes already implemented here look good to me.

Bisaloo:
I fixed the merge conflict so feel free to merge this whenever you want. I'll take some more time to consider the other points with your answers in mind.

PR #504, opened 2021-06-08
Do not beautify json file

Bisaloo:


sbfnk (commented):
WIth this we should remove all the commits since ede6a32193b7e366c1586f00b1bacc6d634c1330 - do you want to do this within the PR?

PR #511, opened 2021-06-14
Store repo-specific settings in a single top-level config file

Bisaloo:
This is a draft PR since it mainly aims at providing a support for an upcoming discussion.

After spending some time looking at the code and the repository structure and because the end goal is to provide a template that can be re-used for any epidemic in any place, I think the way to go is to use a config file for all repo-specific settings. I talk a little bit about this in this issue in covidHubUtils as well: https://github.com/reichlab/covidHubUtils/issues/240.

All the tools I can think of that aim at providing a general template that can be updated to a user specific needs use this config file system: jekyll/hugo/etc. templates for websites hosted on GitHub, projects from the 'downverse' (pkgdown, bookdown).

The idea is that future users can then copy the template / generate it with a function and only tweak the content of the file `forecasthub.yml` instead of having to change hardcoded settings all over the repo.


Bisaloo:
Workflow question @kathsherratt & @sbfnk:

Should we try to merge a first version of this early so that we can already benefit from it (e.g., if we add hospitalisations, we just have to update in one place) or should we wait until we have a more polished version? If you think we should merge the first changes now, do you have any comments/suggestions?

Bisaloo:
`targets` is also stored in zoltar and we can get it with [`zoltr::targets()`](https://reichlab.io/zoltr/reference/targets.html) once this is set up

sbfnk:
I think this is a great idea and a good idea to merge now. Other things this could contain are:
`week_end_date: Saturday` 
`forecast_week_date: Monday`
both of which we might change in the future. They affect (off the top of my head)
https://github.com/epiforecasts/covid19-forecast-hub-europe-validations/blob/8df69bf0b96bd928a356639dbad809b83ddded27/codebase/covid19.py#L141
https://github.com/epiforecasts/covid19-forecast-hub-europe/blob/4d2886d263a63378db6a6e0d0bf2b3d50a90d6fc/code/ensemble/EuroCOVIDhub/create-weekly-ensemble.R#L16
https://github.com/epiforecasts/covid19-forecast-hub-europe/blob/4d2886d263a63378db6a6e0d0bf2b3d50a90d6fc/code/evaluation/score_models.r#L129
https://github.com/epiforecasts/covid19-forecast-hub-europe/blob/4d2886d263a63378db6a6e0d0bf2b3d50a90d6fc/code/reports/compile-ensemble-report.r#L13
https://github.com/epiforecasts/covid19-forecast-hub-europe/blob/4d2886d263a63378db6a6e0d0bf2b3d50a90d6fc/code/reports/compile-evaluation-reports.r#L13

I'm not sure if the `week_end_date` might be hard-wired into some of the `covidHubUtils` functionality.

code/auto_download/auto-download-lanl-covid19.r
```diff
@@ -9,12 +9,13 @@ library("dplyr")
...
 library("readr")
 library("janitor")
+source(here("code", "config_utils", "get_forecast_targets.R"))
 
 model_name <- "LANL-GrowthRate"
 raw_dir <- file.path(tempdir(), "data-raw", model_name)
 processed_dir <- here::here("data-processed", model_name)
 last_sunday <- floor_date(today(), unit = "week", week_start = 7)
-data_types <- c("cases", "deaths")
+data_types <- get_forecast_targets()
```
sbfnk, line 18: Would this make more sense as `get_hub_config("targets")` and `get_hub_config("ensemble_method")` so that we don't have to maintain a separate accessor function for every config option?

code/evaluation/score_models.r
```diff
@@ -6,6 +6,8 @@ library(tidyr)
 library(lubridate)
 library(here)
 library(readr)
+source(here("code", "config_utils", "get_forecast_targets.R"))
```
sbfnk, line 9: would it make more sense to add this to (a fork of) `covidHubUtils` instead?

sbfnk (approved):

Bisaloo:
From https://github.com/epiforecasts/covid19-forecast-hub-europe/pull/105/files#r659991450: add repo url

code/evaluation/score_models.r
Bisaloo: My vote is to merge this first and then, I will submit a PR to covidHubUtils

kathsherratt:
This is really neat @Bisaloo and definitely needed!

These don't need to be priorities for now (don't stop from merging this PR), but I wanted to note a couple of thoughts for expanding this in future -

1. A hub config could be super useful for the validation routine. The validation runs fairly independently once the set of targets, quantiles, and dates (as in Seb's comment) have been defined. So there is definitely potential to future proof the validation code if these definitions are all in the hub config (we could work together on this).

2. The config file also seems quite similar to Zoltar's use of a [project config file](https://docs.zoltardata.com/fileformats/#project-creation-configuration-json). It might be worth checking out the config specification they have (and that we have in [our project](https://www.zoltardata.com/project/238)) for ideas on what a central hub config file could include. (Expanded explanations of the keys [here](https://docs.zoltardata.com/concepts/)). Potentially this would also mean we could feed one to the other, e.g. in setting up Zoltar automatically when creating a new hub, though this would need a bit more work.

code/evaluation/score_models.r
sbfnk: OK

Bisaloo:
> A hub config could be super useful for the validation routine. The validation runs fairly independently once the set of targets, quantiles, and dates (as in Seb's comment) have been defined. So there is definitely potential to future proof the validation code if these definitions are all in the hub config (we could work together on this).

Once this is merged, I can start a branch using the settings added here in the validations. I agree it would be great to add more settings after!

Bisaloo:
I think this is ready to merge but since I've made some additional changes, I requested a new round of review just in case.

I had to force push before there were too many conflicts, sorry for the mess. The last commit from my previous batch of changes is 3d6c287. The new changes start at 538e124.

Bisaloo:
:thinking:, we might need to think about a better solution to get the settings from the yaml file. For example, I would like to add the possible quantiles like this:

```yaml
shortname: EuroCOVIDhub
repo_url: https://github.com/epiforecasts/covid19-forecast-hub-europe
ensemble_method: mean
horizon: '4 weeks'
# week_end_date and forecast_week_date use the ISO 8601 standard (same as the
# lubridate R package).
# 1: Monday, 2: Tuesday, 3: Wednesday, 4: Thursday, 5: Friday, 6: Saturday,
# 7: Sunday
week_end_day: 6
forecast_week_day: 1
target_variables:
  - cases
  - deaths
target_values:
  - point
  - quantiles: [0.01, 0.025, 0.05, 0.10, 0.15, 0.20, 0.25, 0.30, 0.35, 0.40, 0.45, 0.50, 0.55, 0.60, 0.65, 0.70, 0.75, 0.80, 0.85, 0.90, 0.95, 0.975, 0.99]
```

but I cannot find an easy solution to access this information after. And a single function `get_hub_config()` might not be appropriate.

sbfnk:
Isn't there something like `read_yaml` that could be used here? or maybe I'm not understanding what the problem is.

Bisaloo:
All very good points @kathsherratt, thanks!

> It might be worth renaming target_value to type as this how point/quantiles are described elsewhere in the hub

What about something like `target_type` or `forecast_type`? `type` by itself is not self-explanatory when reading the yaml file with no context.

Bisaloo:
> Since the week start might vary, might it maybe be clearer to define the week_end_day / forecast_week_day as the names of the week days (Saturday, Monday)? Then when they are used they can be converted to numeric as needed. I know the numeric is the ISO standard but might still help just to avoid all doubt.

Do you see this conversion taking place in `get_hub_config()` or in downstream scripts?

kathsherratt:
> type by itself is not self-explanatory when reading the yaml file with no context.

Yeah definitely agree. `forecast_type` would work well!

> Do you see this conversion taking place in get_hub_config() or in downstream scripts?

I had assumed downstream but I'm really not tied to any of this. Just looking from a user's perspective asking `get_hub_config()` "What day does the forecast week end?", my sense is the reply which leaves no room for error would be "Saturday". I think the situation of needing a numeric weekday is less general than the name of the week day... But I am surely overthinking it!

Bisaloo:
Ah no, I just tested and I remember why I used `gsub()`, I needed to remove an extra `s`. I agree it's messy though. I'll try to think of something better.

kathsherratt (approved):

Bisaloo:
I reconsidered and I think it makes more sense / it's more future-proof to use 

```yaml
target_variables:
  - inc case
  - inc death
```

instead of

```yaml
target_variables:
  - cases
  - deaths
```

as the US hub for example is using `cum case`, and `cum death` as additional variables.

It still requires a strange `gsub()` but I'm out of ideas here :shrug:

kathsherratt:
That does make sense 👍  A final thought - I guess we could add `target_type: incident` as a key (and then just gsub the first three letters?). It's important that we don't have cumulative targets as it's a major difference from the US hub, so worth highlighting up front.

I agree with you and Seb though that we should merge this asap and make other additions later :)

Bisaloo:
> A final thought - I guess we could add target_type: incident as a key (and then just gsub the first three letters?). It's important that we don't have cumulative targets as it's a major difference from the US hub, so worth highlighting up front.

I would say it probably adds an extra layer of complexity. To get of the target types, we would have to do something like:

```r
apply(
  expand_grid(
    substr(get_hub_config("target_type"), 1, 3),
    get_hub_config("target_variables")
  ),
  1, paste, collapse = " "
)
```

Additionally, what about the (theoretical) case when you have `cum inc` but not `cum death`?

I think having

```yml
target_variables:
  - cum case
  - cum death
  - inc case
  - inc death
```

in the US hub case already makes it very clear that they have cumulative targets and we don't.

kathsherratt:
Ah yeah of course, as it is now makes the best sense then! All good!

PR #561, opened 2021-06-22
Add De/Pl forecasts

jbracher:


kathsherratt:
Thanks @jbracher !

In future weeks we'll need to figure out a way to round any non-integer forecast values among the DE/PL forecasts, before merging into the Euro hub. I'll have a think if we can do this with a GH action at our end, unless you are planning to update DE/PL hub policy to accept integer values.

PR #570, opened 2021-06-28
Zoltar github action

kathsherratt:
Creates github action for uploading to zoltar.

This action works off code on PR #105, but I've pulled the action out into a separate PR. This is so that we can merge the action and then use it to test PR #105 before merging that (by running the action on the #105 branch).

The action is currently on workflow dispatch only, with the idea to update to run at least once daily after we've tested and merged #105.

### DIFF
--- a/.github/workflows/zoltar-upload.yml
+++ b/.github/workflows/zoltar-upload.yml
@@ -12,7 +12,7 @@ jobs:
     steps:
     - uses: actions/checkout@master
       with:
-        submodules: recursive
+        submodules: recursive # zoltar upload relies on validations submodule
 
     - name: Python set up
       uses: actions/setup-python@v2
--- a/.github/workflows/zoltar-upload.yml
+++ b/.github/workflows/zoltar-upload.yml
@@ -1,0 +1,30 @@
+name: Zoltar upload
+
+on:
+  workflow_dispatch #:
+  # schedule:
+  #  - cron: "0 7 * * *"
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/zoltar-upload.yml
```diff
@@ -0,0 +1,30 @@
...
+  #  - cron: "0 7 * * *"
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@master
+      with:
+        submodules: recursive
```
Bisaloo, line 15: Do we need to pull the validations submodule for this?

kathsherratt: Yes as it is at the moment (simple adaptation from the US hub). The process of getting the csv files into a zoltar-ready json relies on the `quantile_io` and `covid19` modules in the validation routine. It would definitely be good to make this simpler in the long term though as it's quite a hassle and not very easy to set up for e.g. a new hub. (Though some of this is probably my lack of python experience as the main hurdle was getting different file structures to work together)

Bisaloo: Ok :+1:, could you please add this as a comment to avoid any future mistake (since we might not always dig through the history of PRs)?

Bisaloo (approved):
LGTM, thanks!

.github/workflows/zoltar-upload.yml
kathsherratt: Good idea, thanks!

kathsherratt:
If all okay with this I will merge, so that we can do a test run using PR #105 to upload the new forecasts this week.

PR #585, opened 2021-06-28
Add files via upload

miraanto:


kathsherratt:
Hi @miraanto , thanks for submitting! Unfortunately there is a very tiny formatting problem which is making the validation checks fail - the team name in the file name is in lowercase, where the team-model folder has an uppercase letter U. 
These need to match to keep everything consistent and make sure we can find your forecasts in future! Please could you just update the file name to match?

Thanks!
Kath

PR #597, opened 2021-06-30
`get_hub_config()` returns all settings by default

kathsherratt:
Just to make it a little easier, adds a default to return all settings from `get_hub_config()` (i.e. the whole yaml).


Bisaloo (approved):
Good idea!

Issue #537, opened 2021-06-17
New forecast format: integer values


kathsherratt:
We are updating the hub forecast format to ensure that **all forecast values are integer counts**. 

This is new as we have not differentiated between integer and non-integer forecasts before. However we think forecasting whole counts makes sense given the forecast targets for cases and/or deaths. It also makes the hub more storage-efficient, which helps keep the project sustainable in the long term.

This will impact all forecast submitted to the hub.

### New submissions
- When formatting forecasts before submission, **please round values to an integer**
- You will also now see a warning message added to forecast submissions which have non-integer values 

### Existing forecasts
- We will be **updating all forecasts in the hub repo** to round all values to integers
- By default, we will use the `round()` function in `R` (where if the decimal is .5, the value is rounded to the nearest even integer)
- We aim to do this for all forecasts by **30 June**: please get in touch before then if you have any concerns

kathsherratt:
Tagging hub contributors:
@jbracher @miraanto @BIOCOM-SC @RobertWalraven @gcgibson @seabbs @dwolffram @cyberklezmer @nikosbosse @aniruddhadiga @Barb-00 @Bisaloo @gr3gr3 @storaged @istirodiah @FrostXTJ @jlaznarte @pmontman @h-veronika @stefanheyder @hgurung-iem @sangeetabhatia03 @TaoSunVoyage @JADEUSC @dmorinya @kraus-stat @semohr @neele-itwm @holgerman @scc-usc

neele-itwm:
Hi Kath,
Can you please confirm whether this applies also to all quantile-forecast values?
Thanks,
Neele

kathsherratt:
Hi Neele - yes, this applies to all point and quantile forecasts. It's fine for multiple quantiles to have the same forecast values, as long as values are either equal or increase with increasing quantiles. 

For example, a forecast for quantiles 0.01 to 0.15 for one target/location could look like:

| forecast_date | target              | target_end_date | location | type     | quantile | value |
|---------------|---------------------|-----------------|----------|----------|----------|-------|
| 2021-06-21    | 1 wk ahead inc case | 2021-06-26      | AT       | quantile | 0.010     | 1     |
| 2021-06-21    | 1 wk ahead inc case | 2021-06-26      | AT       | quantile | 0.025    | 1     |
| 2021-06-21    | 1 wk ahead inc case | 2021-06-26      | AT       | quantile | 0.050     | 2     |
[...]

PR #600, opened 2021-07-01
Round all forecasts

kathsherratt:
Round all existing forecasts in `data-processed` .

See: #537


sbfnk (approved):
Looks good. Have you done a run of e.g. `score_models.r` or re-creating the ensembles to double check this hasn't broken anything (not that I think it has, but just to be safe).

code/validation/round-forecasts.R
```diff
@@ -0,0 +1,12 @@
+# Round all forecasts in data-processed
+library(here)
+library(purrr)
+
+# Load forecast from individual file, round values, and re-write
+all_files <- dir(here("data-processed"), recursive = TRUE,
```
Bisaloo, line 6: It would be slightly more robust to use `dir(..., full.names = TRUE)` so you don't have to re-use `here("data-processed")` after.

This way, if we ever change the location / name of the folder containing the forecast, we just have to edit one line.

It will return absolute file paths but we don't really care since it's just an intermediary step that is never saved or displayed :shrug:

kathsherratt:
@sbfnk , good plan, yes - just checked both ensembling and evaluation and all runs fine.

code/validation/round-forecasts.R
Bisaloo: I think you can drop the `here()` entirely in `read/write_csv()` :wink:

PR #598, opened 2021-06-30
Add baseline model

Bisaloo:
Should this baseline model taken into account when building the ensemble or should `create-baseline.R` automatically add it to the list of exclusions each week?

Fix #531

PR #602, opened 2021-07-01
Exclude models designated `other`

kathsherratt:
Models with `team_model_designation: other` in their metadata should be excluded from the ensemble.

This PR adds the option  `exclude_designated_other` to `use_ensemble_criteria()`, defaulting to `TRUE`.

This arg is added to all other ensemble code and set to default TRUE everywhere.

This also means that past ensembles should not have included the `other`-designated models:
- `epiforecasts-EpiExpert_Rt` 
- `epiforecasts-EpiExpert_direct`

Which affects forecasts for DE, PL, NL, and GB. 

@sbfnk could you confirm you're happy for me to update the past ensembles (I will put in a separate PR)?


kathsherratt:
This supports #598 if baseline model is designated `other`

code/ensemble/utils/run-ensemble.R
```diff
@@ -35,6 +35,7 @@ source(here("code", "config_utils", "get_hub_config.R"))
 run_ensemble <- function(method = "mean",
                          forecast_date,
                          exclude_models = NULL,
+                         exclude_designated_other = TRUE,
```
sbfnk, line 38: Do you think it's worth having this as an explicit option here? An alternative would be to have `...` here and pass `...` to `use_ensemble_criteria` instead.

code/ensemble/utils/run-multiple-ensembles.R
```diff
@@ -18,6 +18,7 @@ source(here("code", "ensemble", "utils", "run-ensemble.R"))
 run_multiple_ensembles <- function(forecast_dates,
                                    methods,
                                    exclude_models = NULL,
+                                   exclude_designated_other = TRUE,
```
sbfnk, line 21: I think this function only needs `forecast_dates` and `methods` as arguments. Everything else could be passed via `...`

```
run_multiple_ensembles <- function(forecast_dates, 
                                   methods, 
                                   ...) {
```

and later

```
  ensembles <- map2(methods,
                    dates,
                    ~ safe_run_ensemble(method = .x,
                                        forecast_date = .y,
                                        ...)
```

code/ensemble/utils/use-ensemble-criteria.R
```diff
@@ -51,7 +53,15 @@ use_ensemble_criteria <- function(forecasts,
               by = c("model", "target_variable", "location")) %>%
     mutate(not_excluded_manually = !(model %in% exclude_models)) %>%
   # 4. Drop hub ensemble model
-    filter(!grepl("EuroCOVIDhub", model))
+    filter(!grepl("EuroCOVIDhub-ensemble", model))
```
sbfnk, line 56: good catch!

sbfnk (changes requested):

code/ensemble/utils/run-multiple-ensembles.R
kathsherratt: Ah, yes that would be better - thank you!

code/ensemble/methods/create-ensemble-relative-skill.R
```diff
@@ -37,9 +37,14 @@ create_ensemble_relative_skill <- function(forecasts,
                                            average = "mean",
                                            by_horizon = FALSE,
                                            return_criteria = FALSE,
-                                           verbose = FALSE) {
+                                           verbose = FALSE,
+                                           ...) {
```
sbfnk, line 41: I might be missing something but I don't think this one is needed.

code/ensemble/utils/use-ensemble-criteria.R
```diff
@@ -19,7 +19,8 @@ source(here("code", "ensemble", "utils", "get_model_designations.r"))
 use_ensemble_criteria <- function(forecasts,
                                   exclude_models = NULL,
                                   exclude_designated_other = TRUE,
-                                  return_criteria = TRUE) {
+                                  return_criteria = TRUE,
+                                  ...) {
```
sbfnk, line 23: I don't think this one is needed

code/ensemble/utils/use-ensemble-criteria.R
```diff
@@ -57,7 +58,7 @@ use_ensemble_criteria <- function(forecasts,
   
   # 5. Drop "other" designated models
   if (exclude_designated_other) {
-    not_other <- get_model_designations(here()) %>%
+    not_other <- suppressWarnings(get_model_designations(here())) %>%
```
sbfnk, line 61: Any particular reason we want to suppress warnings? Just thinking that we might sometimes get something informative about problems with the yaml files that could get lost here.

sbfnk (approved):

code/ensemble/utils/use-ensemble-criteria.R
kathsherratt: Fair, I was just seeing lots of line ending warnings which didn't have any impact, but I'll remove.

code/ensemble/methods/create-ensemble-relative-skill.R
kathsherratt: I've tried both ways but without the `...` I get an `unused argument` error which annoyingly stops the whole lot from running.

sbfnk: To be honest this points to a problem to me as there I'm not clear which argument could be passed here that's known to `create_ensemble_relative_skill`. I'm probably missing something obvious but do you remember which argument the error was about?

kathsherratt: The unused argument was `exclude_designated_other`, which gets passed to `create_ensemble_relative_skill()` whenever it's run inside the wrapper of `run_ensemble(method = "relative_skill_")`.

`run_ensemble()` should include a way to pass on the necessary arguments for relative skill (`evaluation_date` and `continuous_weeks`), but I also think we shouldn't be forced to provide the args, because there are sensible defaults in `create_ensemble_relative_skill()` if they're missing. 

This can be achieved by passing along `...`, but that does then mean that  `create_ensemble_relative_skill()` has to accept all of whatever other general arguments are needed for the rest of `run_ensemble()`, as well as the specific args for this function. Hence if `create_ensemble_relative_skill()` doesn't have the `...`, I was getting the `unused argument` error.

I guess the alternative to really avoid the `...` in `create_ensemble_relative_skill()`, would be to grab the defaults for any missing args inside `run_ensemble()` before before it's run - like:

```
run_ensemble <- function(...) {
 extra_args <- list(...)
 # If necessary arguments not passed in through ... args, get function's defaults
    if (!"evaluation_date" %in% names(extra_args)) {
      evaluation_date = formals(create_ensemble_relative_skill)$evaluation_date
      }
 ensemble <- create_ensemble_relative_skill(evaluation_date)
```
Although this seems a bit more prone to failing if arguments change.

I might also be massively over-thinking this. I was trying to keep the ensemble code flexible so put it in this kind of nesting structure, but that may well have just over complicated it.

sbfnk: Had a go at a simpler solution in https://github.com/epiforecasts/covid19-forecast-hub-europe/pull/609

PR #603, opened 2021-07-01
Exclude `other` designated models from past ensembles

kathsherratt:
Works from #602 to re-create all past ensembles without models designated `other`. Used: https://github.com/epiforecasts/covid19-forecast-hub-europe/blob/main/code/ensemble/EuroCOVIDhub/re-create-weekly-ensembles.R

Note that between updates the `Priesmann-Bayes` model was renamed to `DSMPG-Bayes` so this created a diff in the inclusion criteria, but no substantial change to model inclusion.

Also updates all evaluation scores using score_models.r with


sbfnk (approved):
Thanks Kath, this looks good. I noticed that there are a few other models that got added along the way on some dates. I presume these were late submissions where we didn't decide to exclude them but also didn't re-run the ensemble - happy for them to be included now.

kathsherratt:
Ah, thank you for flagging! Just to be sure / for completeness I checked the added forecasts:
- `UB-BSLCoV` on 2021-06-07 - it was committed together with the following week's forecast [here](https://github.com/epiforecasts/covid19-forecast-hub-europe/pull/508)
- `LANL-GrowthRate` on 2021-06-14 - committed on the Wednesday (June 16): our gh action must have run late I think.

Anyway, before doing anything here I will merge #600 , then re-run this PR (from #602) so that the ensembles / evaluation are updated with the rounded forecasts.

sbfnk:
Great. Worth merging `main` as well before re-creating the ensembles as I just added a missing [LANL forecast](https://github.com/epiforecasts/covid19-forecast-hub-europe/commit/7bb522953916b3aa2be553eac9919f2438c716b2).

kathsherratt:
I've updated so that past ensembles reflect the rounding of forecasts and some missing forecasts added back in.

I've also updated the evaluation, just to keep the scores correct, although this will presumably be very soon over-written by #614; and possibly affected also by the rounding discussion in #638.

sbfnk:
That's a good point - probably best to remove the evaluations for now so as to not grow the repo unnecessarily.

PR #606, opened 2021-07-02
Submission 2021-07-05

h-veronika:


h-veronika:
Added under 20 age compartment for probability of hospitalization and death estimate.

PR #609, opened 2021-07-02
trying to update ensemble options to be more robust

sbfnk:
Linked to [this discussion](https://github.com/epiforecasts/covid19-forecast-hub-europe/pull/602#discussion_r663084918). I think it's fine to set the default in `run_ensemble` - we want all the function that (re-)create ensembles to use this default anyway, and don't want to rely on remembering to set this explicitly, I think.


kathsherratt:
Ah looks good, sorry if that was a bit of a mess on my part. Thank you Seb!

kathsherratt (approved):

PR #613, opened 2021-07-05
ILM-EKF forecast for 2021-07-05

stefanheyder:


kathsherratt:
Hi @stefanheyder - I noticed this PR adds a duplicate of your team's forecast from [last week](https://github.com/epiforecasts/covid19-forecast-hub-europe/commit/2ab02edca54c9ca2838c6b35d5cf6c2321b4fdde). (I think this might have been from creating the PR from a different origin last week). Obviously the forecasts are duplicates and it will be an identical replacement so it's not a problem at all. I only noticed since we check the commit history of any past forecasts. Just adding this comment for the record before merging!

PR #620, opened 2021-07-05
Add hospital data

kathsherratt:
Adds daily hospital data from ECDC  to existing ECDC download script
- Data pulled from ECDC hosted source
- Access to source requires environment variables (/ github secrets)
- Filtered for data from a public source: one of Github API, Country API, Country website
- Formatted and saved in hub standard
- Data should be accessible using `covidHubUtils::load_truth()` - this just needs a quick PR into the package

Modifies Github action to support this:
- Data download run in existing github action
- Action updated to run daily
- Uses three additional secrets set in the github action

I also noticed some out of date code in the data download for ECDC cases and deaths (still using ISO-3 country codes, and we've found readr more reliable than vroom) so made those minor changes.

### DIFF
--- a/.github/workflows/ECDC.yml
+++ b/.github/workflows/ECDC.yml
@@ -2,7 +2,7 @@ name: "ECDC"
 on:
   workflow_dispatch:
   schedule:
-    - cron: "0 12 * * 4"
+    - cron: "0 12 * * *"
 
 jobs:
   get_ecdc:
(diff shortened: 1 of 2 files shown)


sbfnk (approved):
Looks good to me. I'm not convinced of having a double naming convention where a file name `Incident Hospitalizations` corresponds to a `target_variable` called `inc hosp` but I realise we have inherited this from the US hub, and suspect that changing it might break things.

sbfnk:
PS: perhaps this will all be fixed once we can add truth data to Zoltar, as relying on data sources with hard-coded file names in `covidHubUtils` might not be the best solution anyway

kathsherratt:
Completely agree, I'd prefer not to rely on the `covidHubUtils::load_truth()` from local repo - I'll prioritise getting truth data on to Zoltar.

kathsherratt:
Links with `covidHubUtils`: https://github.com/reichlab/covidHubUtils/pull/257

PR #658, opened 2021-07-12
better filtering for last X forecasts

sbfnk:
At the moment filtering for the last X forecasts is done with `head`. This can lead to gaps when there are missing dates due to data issues, e.g. https://covid19forecasthub.eu/reports/evaluation-report-Ireland.html

This aims to fix this issue.


Bisaloo (approved):

Bisaloo:
I think this created a problem for France: https://covid19forecasthub.eu/reports/evaluation-report-France.html

I'll investigate

sbfnk:
I think this was actually there before - I noticed it earlier this week but couldn't remember which country it was in and hoped it would be fixed by this.

Bisaloo:
Ah no, you're right. This seems to fix it (at least locally). It's just that the evaluation reports were not re-created since this PR got merged.

PR #673, opened 2021-07-14
Set fig height dynamically

### LINKED ISSUE #562: Use scales = "free_y" for facetted plot
see e.g.:

image

I think it is a bit unintuitive to read because models with different horizons don't have the same x-axis. It can easily be fixed by specifying `scales = "free_y"` instad of `scales = "free"`.

To avoid endless git conflicts, I'd like to wait until I'm done with #469 before changing this if you agree with this change.

### LINKED ISSUE #671: ensemble plots need to have greater `fig.height`
The plot at https://covid19forecasthub.eu/reports/ensemble-report.html is squashed. It used to have `fig.height=14` (https://github.com/epiforecasts/covid19-forecast-hub-europe/commit/e7dde7f1bed33fb1be8433f8978dc3c8f488cf4d#diff-f65330c9ef062738afa7ab50548a7c4949f550eb6e5802aa1565d85e3f5b7441) but now 8 like the other forecast plots.

I guess ideally the height would be set dynamically based on the number of rows (number of models in the evaluation reports, number of countries in the ensemble report) but otherwise can we set this back to 14 somehow?

Bisaloo:
Fix #562 and fix #671

### DIFF
--- a/code/reports/rmdchunks/plot-forecasts.Rmd
+++ b/code/reports/rmdchunks/plot-forecasts.Rmd
@@ -40,7 +40,9 @@ for (forecast_date in as.character(forecast_dates)) {
         ggplot2::theme(legend.position = "bottom", 
                        strip.placement = "outside") + 
         scale_y_continuous(labels = scales::comma) + 
-        expand_limits(y = 0)
+        expand_limits(y = 0) +
+        # Make sure negative values for cases/deaths are not displayed
+        coord_cartesian(ylim = c(0, NA))
 
       if (is.null(plot)) {
         cat("No forecasts available, possibly because of recent anomalies in the available data." )
--- a/code/reports/rmdchunks/plot-forecasts.Rmd
+++ b/code/reports/rmdchunks/plot-forecasts.Rmd
@@ -15,7 +15,7 @@ forecast_dates <- data %>%
   pull(forecast_date)
 ```
 
-```{r prediction-plots, echo = FALSE, results='asis', fig.width = 8.5, fig.height=8}
+```{r prediction-plots, echo = FALSE, results='asis', fig.width = 8.5, fig.asp = length(unique(rlang::eval_tidy(facetting_var, data = data))) / 16}
 for (forecast_date in as.character(forecast_dates)) {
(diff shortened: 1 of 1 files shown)


code/reports/rmdchunks/plot-forecasts.Rmd
```diff
@@ -36,12 +36,11 @@ for (forecast_date in as.character(forecast_dates)) {
         facet_formula =  rlang::expr(~ !!facetting_var),
         ncol = 3,
         allow_truth_without_pred = FALSE,
-        scales = "free") + 
+        scales = "free_y") + 
         ggplot2::theme(legend.position = "bottom", 
                        strip.placement = "outside") + 
         scale_y_continuous(labels = scales::comma) + 
-        expand_limits(y = 0) + 
-        coord_cartesian(ylim = c(0, NA))
```
Bisaloo, line 44: I removed this because it doesn't seem to serve any purpose but I might be wrong. Do you have any idea why it might be necessary?

sbfnk: Yes, we sometimes end up with negative values because the incident case/death numbers are based on diffs of cumulative numbers, and sometimes these go down if e.g. duplicates are removed etc. In that case I think we want to not show the negative numbers.

code/reports/rmdchunks/plot-forecasts.Rmd
```diff
@@ -15,7 +15,7 @@ forecast_dates <- data %>%
   pull(forecast_date)
 ```
 
-```{r prediction-plots, echo = FALSE, results='asis', fig.width = 8.5, fig.height=8}
+```{r prediction-plots, echo = FALSE, results='asis', fig.width = 8.5, fig.asp = length(unique(rlang::eval_tidy(facetting_var, data = data))) / 16}
```
sbfnk, line 18: Cool! No idea what this will look like but I trust you've chosen something sensible.

sbfnk (approved):

Bisaloo:
Example for Germany:

image

Example for ensemble:

image

sbfnk:
Looks good!

PR #674, opened 2021-07-14
Update x axis in plots

sbfnk:
Trying to go for something more human-readable.


code/reports/rmdchunks/plot-forecasts.Rmd
```diff
@@ -52,7 +52,8 @@ for (forecast_date in as.character(forecast_dates)) {
         scale_y_continuous(labels = scales::comma) + 
         expand_limits(y = 0) +
         # Make sure negative values for cases/deaths are not displayed
-        coord_cartesian(ylim = c(0, NA))
+        coord_cartesian(ylim = c(0, NA)) +
+        xlab("Last day of week")
```
Bisaloo, line 56: What about just "week"? It's admittedly a little less precise but it would be easier to understand IMO.

PR #676, opened 2021-07-15
fix LANL autodownload script

sbfnk:
Fixes problem in [action](https://github.com/epiforecasts/covid19-forecast-hub-europe/actions/runs/1032155817) introduced by adding a hospitalisation target in https://github.com/epiforecasts/covid19-forecast-hub-europe/pull/672

Possibly simpler alternative would be to explicitly loop over only cases and deaths here, avoiding the failed http request.


Bisaloo:
In this case, I guess it makes sense to say explicitly that we want deaths and cases instead of getting it from the config file. These are properties of the source (LANL), not of the hub.

PR #677, opened 2021-07-16
Use scaled_rel_skill instead of relative_skill

Bisaloo:
Missed comment from https://github.com/epiforecasts/covid19-forecast-hub-europe/pull/635#issuecomment-878358535


sbfnk (approved):
good catch

PR #635, opened 2021-07-06
Score forecasts relative to baseline

Bisaloo:
Fix #614 

~~I just created the previous baselines for now. I'm waiting for #577 to be merged before making any more changes to the reports to avoid conflicts.~~

PR #682, opened 2021-07-19
update forecast reports

sbfnk:
Visible changes:
- updates forecast report table according to suggestions in https://github.com/epiforecasts/covid19-forecast-hub-europe/issues/594

Under the hood:
- implements suggestion in https://github.com/epiforecasts/covid19-forecast-hub-europe/issues/595. Forecasts are now evaluated once when generating multiple reports, improving runtime
- uses a single report structure for the overall and country-by-country reports 
- saves evaluations, rendering https://github.com/epiforecasts/covid19-forecast-hub-europe/tree/main/code/evaluation obsolete (addressing https://github.com/epiforecasts/covid19-forecast-hub-europe/issues/469)

Updated table columns:
image


Bisaloo (commented):
I'll review more carefully later today but before I forget: the file `score_models.r` and the corresponding GitHub action (`evaluation.yml`) should probably be deleted as well.

On a related note, we need to think about the impact on the workflow when we have to re-run all evaluations (as in #635).

code/reports/rmdchunks/ranking-table.Rmd
```diff
@@ -1,111 +1,25 @@
...
-            "interval_score", "sharpness",
-            "underprediction", "overprediction",
-            "bias", "aem")
-
-  table %>%
-    select(all_of(cols)) %>%
-    arrange(scaled_rel_skill) %>%
-    mutate_if(is.numeric, round, 2) %>%
+  df %>%
+    select(model, n, n_loc, rel_wis, rel_ae, cov_50, cov_95, bias) %>%
```
Bisaloo, line 18: :grey_question: 

`n_loc` is always 1 in the case of a country and unnecessary columns add extra cognitive load. Should we treat this differently when `location_name = "Overall"` vs the other cases?

code/reports/rmdchunks/filter-forecasts.Rmd
```diff
@@ -0,0 +1,22 @@
...
+so that we can cache the loading and scoring step.. -->
+```{r}
+if (params$location_name != "Overall") {
+  data <- data %>%
+    filter(location_name == params$location_name)
+}
+table <- table %>%
+  filter(location_name == params$location_name)
+
+data <- data %>%
```
Bisaloo, line 11: :+1:

Nice idea to filter this here!

code/reports/rmdchunks/save-scores.Rmd
```diff
@@ -0,0 +1,7 @@
+<!--- Save evaluation scores as csv -->
+```{r, cache = TRUE, cache.extra = params$report_date}
```
Bisaloo, line 2: :red_circle: 

I don't think this should be cached(?)

code/reports/rmdchunks/score-forecasts.Rmd
```diff
@@ -0,0 +1,152 @@
...
+                  )) %>%
+  left_join(num_loc, by = c(
+                       "model", "target_variable", "horizon",
+                       "location"
+                     )) %>%
+  left_join(num_fc, by = c(
+                      "model", "target_variable", "horizon",
+                      "location"
+                    )) %>%
+  left_join(locations, by = "location") %>%
```
Bisaloo, line 146: :grey_question: 

inor point: would it make sense to use the countrycode package here? It adds an extra dependency but it would make the intent much clearer IMO. At the beginning of this chunk, I was wondering why we need to save all the locations.

Bisaloo (changes requested):
Overall very useful change! I think it's really going in the right direction :100:.

Trying something new in my review. I marked as:
- :red_circle: bugs or possible bugs that should be addressed before merging
- :grey_question: suggestions of alternative implementation but feel free to disagree / reject my suggestions
- :+1: just an approving comment that requires no specific action on your end

code/reports/rmdchunks/score-forecasts.Rmd
nikosbosse: I think that the country codes are also exported from covidHubUtils - so we could use those actually.

code/reports/evaluation/evaluation-report.Rmd
```diff
@@ -37,19 +41,39 @@ include_countries <- include_forecast_plot ||
...
 data_types <- get_hub_config("target_variables")
+quantiles <- get_hub_config("forecast_type")$quantiles
 ```
 
 <!--- Chunk to load forecast and truth data and consolidate it in a single df called data. Usually called as the first chunk. This step is cached so make sure to delete the cached files if necessary when testing. -->
 ```{r load-data, child = "../rmdchunks/load-data.Rmd"}
 ```
 
+<!--- Evaluate and score forecasts -->
+```{r score-forecasts, child = "../rmdchunks/score-forecasts.Rmd"}
```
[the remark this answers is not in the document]
sbfnk: My inclination was to separate evaluation and report generation (though ultimately not really achieved here) - one is doing the computation and the other one drawing pictures and tables. I don't really see there being a "computing again" step. I don't feel particularly strongly on this issue though - maybe we just have different views on what is the "data" is that is being processed here (raw forecasts vs. forecast scores).

Worth bearing in mind that we do want to pull everything together to put in a csv every week (to be used for the trained ensemble) so if we load->filter->compute we'll have to either somehow bind the computed scores before saving to csv, or maintain a separate bit of code that scores and saves to csv, as it is at the moment (meaning we'll do the scoring twice every week).

Ultimately this should be determined by what is most easy to maintain as code base, and easiest generalised. Happy for you to make that call.

code/reports/rmdchunks/ranking-table.Rmd
sbfnk: Yes, I think this makes sense. We then just need to also have a conditional in the description, (which isn't a problem though)
https://github.com/epiforecasts/covid19-forecast-hub-europe/blob/d69c4b3a49fa23b3c7db048db2db63b153261a2a/code/reports/rmdchunks/ranking.Rmd#L31

code/reports/rmdchunks/save-scores.Rmd
sbfnk: This was to prevent the `write_csv` call if there was no update.

code/reports/rmdchunks/score-forecasts.Rmd
sbfnk: I'm not sure either is a good idea. What if we want to run this on things that aren't the ECDC hub, or aren't countries (e.g. investigate our US forecasts that are state-by-state)? My inclination would be to make the chunks as self-contained as possible (i.e., get information from the existing tables if already present, rather than from outside packages), but perhaps I'm overemphasising the desire for this to be generalisable, at the expense of code stability/readability.

code/reports/evaluation/evaluation-report.Rmd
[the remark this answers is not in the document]
Bisaloo: > Worth bearing in mind that we do want to pull everything together to put in a csv every week (to be used for the trained ensemble) so if we load->filter->compute we'll have to either somehow bind the computed scores before saving to csv, or maintain a separate bit of code that scores and saves to csv, as it is at the moment (meaning we'll do the scoring twice every week).

Good point! Let's go with your current version then :+1:

code/reports/rmdchunks/save-scores.Rmd
Bisaloo: As it's not a computationally intensive, I'd rather do this step again even if it's not necessary (which shouldn't be happen in the normal use case). 

[As the developer of knitr keeps repeating](https://yihui.org/en/2018/06/cache-invalidation/), caching and caching invalidation are tricky and it's easy to get bitten with something that doesn't get updated when it should.

In this case for example, I **think** we don't invalidate the cache often enough because any change in `table` (e.g., if we changed the chunks above) will not trigger a re-run of this chunk. To get the desired behaviour, we would need to add another dependency to `table` in `cache.extra` or `autodep = TRUE` ([the default seems to be `FALSE`](https://yihui.org/knitr/options/#cache)).

code/reports/rmdchunks/score-forecasts.Rmd
Bisaloo: Maybe the question is more genera then (and we might discuss it in another issue). Do we really need to have both `location_name` and `location` or can we avoid the pain of merging to get one from the other in many instances throughout the hub code?

code/reports/rmdchunks/save-scores.Rmd
sbfnk: This makes sense - happy to remove.

code/reports/rmdchunks/score-forecasts.Rmd
Bisaloo: I'll answer my own comment: yes, we probably need both `location_name` and `location` in the case of other spatial cases. It's a bit clunky for us because there is no real ambiguity if we worked just the country names but if we think about regions or municipalities, it's much more robust and maybe even necessary to use codes.

code/reports/evaluation/evaluation-report.Rmd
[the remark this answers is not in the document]
sbfnk: OK. Very happy to have this revisited at a later stage.

code/reports/rmdchunks/score-forecasts.Rmd
sbfnk: I'm not sure actually - it seems to me that `location_name` is superfluous but given that teams are asked to put it into the forecast submissions (and I think it's being validated) we may as well use it.

code/reports/rmdchunks/ranking-table.Rmd
sbfnk: addressed in c5cb51f and b532f7a

code/reports/rmdchunks/save-scores.Rmd
sbfnk: addressed in 9b293b6

Bisaloo (approved):

PR #683, opened 2021-07-19
Package R code 

Bisaloo:
Draft because #682 will cause serious conflicts so it makes more sense to delay any review until #682 is merged.


sbfnk:
Makes sense - would this create issues with reading the config file though?

kathsherratt:
I generally would prefer to keep all our internal hub code separate from the external hub data (submissions etc) so moving the package to a separate repo would be great. 

Although could this potentially also create problems with running/saving the GH action workflows? 😕

Bisaloo:
> Makes sense - would this create issues with reading the config file though?

No, it doesn't matter where the function `get_hub_config()` is defined to find the config file. It only matters where it is called.

> Although could this potentially also create problems with running/saving the GH action workflows? confused

None that I could think of. The package will be installed via `remotes::install_github()` in the actions, in the same way covidHubUtils currently is. Do you have a specific action / problem in mind?

Bisaloo:
/rebase because I want to run more tests locally and I need the fix from 7a8fa47a69ce4a6b0479ab14066f92d14379bbd0.

Bisaloo:
Let's try to /rebase again

sbfnk:
I've slightly lost track of this - is it ready to merge?

Bisaloo:
Yeah, it should be. I'll be more comfortable if it gets another round of review though. Since some modifications have been made since the last reviews. (I'll review it myself as well since I'm not sure I integrated all the changes done on `main` since then.)

Bisaloo:
I've had a look at this and I think it's ready for another round of reviews whenever you get a moment.

sbfnk (approved):
This all looks good to me. Two suggestions (not necessarily for immediate implementation):

- maybe it would be good to have a general "load forecasts, load truth data, exclude whatever needs excluding for whatever reason and return a merged data frame" function, i.e. what's currently duplicated in `load-data.Rmd` and `score-models.R`
- also maybe `score_models` in `EuroForecastHub` could be slightly more general and have 1) the inclusion criteria for "Overall" (1/2 of locations) as some sort of argument and 2) an option to return e.g. un-averaged scores by "target_end_date" (so that one could use it to get all the 'raw' scores). Perhaps the `summarise_by` argument to all the calls to `eval_forecasts` in `EuroForecastHub::score_models` could be an argument to `score_models` itself?

PR #706, opened 2021-07-20
Add `validation` to `gitignore`

kathsherratt:
I'd like to add the `validation` folder to gitignore. It's a submodule in any case so shouldn't be changed from the main repo, so this should prevent accidental edits (or out of date versions of the submodule) being pushed alongside other commits.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ env
 
 # HTML-rendered version of READMEs
 README.html
+validation
--- a/.gitignore
+++ b/.gitignore
@@ -23,4 +23,3 @@ env
 
 # HTML-rendered version of READMEs
 README.html
-validation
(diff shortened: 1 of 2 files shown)


Bisaloo:
Good idea! After reading the docs, it looks like the proper way to do this is to add 

	ignore = all

at the end of our `.gitmodules` file.

It seems to work locally for me, could you try to see if everything is okay on your end if you do this change? 

Please let me know if I'm not clear and you'd like me to submit another PR to show what I mean.

kathsherratt:
Thanks Hugo! (Checking the docs, a much better idea! 😄) 

I've done as suggested and works fine for me (changes to validation files don't show up as modified files).

Bisaloo (approved):

PR #710, opened 2021-07-23
Add truth sources to location data

kathsherratt:
Adds to `locations_eu.csv`: columns by each `target_variable` with the source of truth data. 

The source and variables are taken from the defaults returned by `covidHubUtils`.

This is useful because truth source varies by target variable and location:

`inc case` and `inc death`
- all locations  = `JHU`

`inc hosp`
- some locations have data = `ECDC`
- some locations have no truth data = `NA`

This info is needed for validating `inc hosp` forecasts, since we now have separate locations valid for each target (this will be a separate PR to validations repo).

I've checked across our code base where I can find `locations_eu.csv` in use (LANL, ECDC, JHU data downloads; data-truth README; visualisation workflow; and in the validations). All runs fine - the csv already had population data so all code handles extra columns already.


sbfnk (commented):
It makes sense but I wonder if there is a more elegant solution that doesn't require us to keep this file updated.

E.g. could we make the entries in https://github.com/epiforecasts/covid19-forecast-hub-europe/blob/main/forecasthub.yml something like

```
target_variables:
  inc case:
    source: JHU
  inc death:
    source: JHU
  inc hosp:
    source: ECDC
```

And get the list of countries from the latest corresponding data file in `data-truth`?

cc @Bisaloo in case he has an opinion on (ab-)use of the config file for this.

kathsherratt:
I agree it would be neat to have the source in the config file. 

I started that way, but then realised it would cause a little extra trouble feeding this through to validating forecasts by a combination of valid location by target variable. Validation already uses `locations_eu.csv` (copied in to the validations repo) but doesn't currently access the truth data in any way. Presumably that could all be managed though. 

Alternatively I guess we could skip all this by just validating `inc hosp` forecasts against all our hub locations (as now). So take hospital submissions for any location, regardless of whether the ECDC currently has public data for it, and accept that they are for now ignored in all evaluation / ensembling since we don't have truth data. That sounds odd and would cost us storage, but then again maybe a positive that we'd already be collecting forecasts if the ECDC (or us) were to add other countries' data at some later date. I don't know if that would happen / is what we would want though.

sbfnk:
Ah, this makes sense. Could the validations repo use `covidHubUtils::load_truth` reading from remote repo perhaps? If not your solution is probably best.

kathsherratt:
That would be ideal, but unfortunately as the validations is still in Python and `covidHubUtils` is an R package, we can't currently get `load_truth()` easily into the validations repo. 

But this will all get simpler when we package up validation in R!

sbfnk:
Of course! The root of all our problems...

sbfnk:
Happy for this to be merged as a solution for now.

PR #733, opened 2021-07-27
Add De/Pl forecasts

jbracher:


jbracher:
Will fix this before noon

PR #735, opened 2021-07-30
Convert prepare_truth_data to R and add hospitalizations

Bisaloo:
We need to add hospitalizations in `viz/truth_to_plot.csv` for the submission app (and for the dashboard on the website but we might also need additional changes for this).

I used this opportunity to convert the script to R (9aa73a1). I checked that it produces the exact same output as the python script.

Then, I added hospitalizations (2be7b9b) and updated the GH action (810287b).

I can also convert the remaining python scripts in this PR if you think it's worth it.


kathsherratt (approved):

kathsherratt:
Great idea, thanks Hugo!

PR #752, opened 2021-08-02
Predictions 02-08-21

BIOCOM-SC:


kathsherratt:
Hi @BIOCOM-SC - thanks for submitting! 

It looks like you have added a duplicate forecast file in the root of the repository, as well as in your team folder. 

I can see a forecast with the filepath `2021-08-02-BIOCOMSC-Gompertz.csv` ; and another forecast saved correctly in your team folder, `data-processed/BIOCOMSC-Gompertz/2021-08-02-BIOCOMSC-Gompertz.csv`.

I've checked that these files are duplicates. 

This is such a small issue that blocks your PR from being merged, that I think it will be quickest if I solve directly in the PR. 

I will add a commit to this PR which deletes the extra file; then I can merge this PR straightaway.

Adding a file to the root instead of the forecast file is a really common issue,  so I'll make sure the automated validation flags this in future!

Kath

Issue #708, opened 2021-07-21
Consider splitting `target` in `target_variable` and `horizon` in forecasts


Bisaloo:
This might not possible for the current iteration of the hub but I think it could be worthwhile to consider it for the next time. The current `target` column contains two pieces of information that would probably make more sense separated.

I started to think about it while trying to loop over `target_type` in the submission app to reduce code duplication:

https://github.com/epiforecasts/ForecastHubSubmissionApp/blob/efeec37c6781adf2349e407380e6cf0f9fb986b5/R/app_server.R#L96-L127

I also intuitively thought this was two different columns when I drafted the specs for the validation package:

> - test that horizons match horizons in the config file
> - test that target types match target types in the config file

Beyond the conceptual argument, this would make our lives easier for the downstream analyses where we always end up splitting this column in 2.

sbfnk:
Very much agree that every column should contain exactly one piece of information: arguably `1 wk ahead inc case` contains 4 (1, wk, inc, wk) and "ahead" is redundant.

Two related questions are:
- which information to include: if `target_end_date` and `forecast date` are given then `1 wk ahead` is redundant but possibly useful for checking that this is interpreted as intended by the forecaster. 
- which information to include in file names and which to include in tables. My inclination would be to be independent of file names but including information there makes it easier to e.g. load only forecasts for certain dates without having to read in all files; it also reduces the amount of repetition affecting storage. Ultimately csv might not be the ideal format for this and it could be that eventually it's best handled by something more suitable like Zoltar or e.g. `json`.

Bisaloo:
On a related note, if we split the columns, it would be worthwhile to spell out entire words: cases/deaths/hospitalizations, incident/cumulative. We always want to full word in visualizations & reports and the current approach of having a dictionary (hosp <-> hospitalizations, case <-> cases, etc.) does not allow for generalization to any target without code changes.

Bisaloo:
It's too late to implement in the forecast hub but we agreed to add it to the standard with the data format working group including the other hubs.

PR #789, opened 2021-08-14
Add hospitalisations to reports

Bisaloo:
I'm really not happy about this fix. The goal would be for everything to work when we change the config file, without having to keep a dictionary with nice labels for plots. The solution would be to use the nice labels directly in the raw data instead of abbreviated terms ('hosp') and to split 'inc case' into 'inc' and 'case'. Related: #708 and #796  

To do before it's ready to merge:

- [x] ~~Check why ensemble doesn't include hospitalisations and fix it.~~ As Seb suggested, we just needed to have a least 4 consecutive weeks of submission for this target.

### DIFF
--- a/code/reports/rmdchunks/get-config-vars.Rmd
+++ b/code/reports/rmdchunks/get-config-vars.Rmd
@@ -10,7 +10,11 @@ countries_hub <- get_hub_config("forecast_locations", config_file = params$confi
 
 # FIXME: this breaks the fact that only changes in the config file should be 
 # necessary to add a new target
-target_variables <- c(Cases = "inc case", Deaths = "inc death")
+target_variables <- c(
+  Cases = "inc case", 
+  Deaths = "inc death",
+  Hospitalisations = "inc hosp"
+)
 ```
 
 ## {.unlisted .unnumbered}


Bisaloo:
This seems to be working fine on my computer. Screenshots below

Screen Shot 2021-09-07 at 14 01 57-fullpage

Screen Shot 2021-09-07 at 14 09 01-fullpage

Bisaloo:
Hmm, noticing just now that hospitalizations are still missing from scores

Bisaloo:
The problem is that data from the ECDC doesn't use the same week origin as our forecast. They use Saturday and we use Monday. So when we merge the forecast and the truth data, they are not matched correctly, as shown in the table below:

```r
knitr::kable(data[10679:10685, ])
```

|location |target_variable |target_end_date |location_name | population| true_value|model                 |forecast_date |horizon |temporal_resolution |type     | quantile| prediction|
|:--------|:---------------|:---------------|:-------------|----------:|----------:|:---------------------|:-------------|:-------|:-------------------|:--------|--------:|----------:|
|BE       |inc hosp        |2021-08-21      |Belgium       |   11372068|        414|NA                    |NA            |NA      |NA                  |NA       |       NA|         NA|
|BE       |inc hosp        |2021-08-28      |Belgium       |   11372068|        419|NA                    |NA            |NA      |NA                  |NA       |       NA|         NA|
|BE       |inc hosp        |2021-09-04      |Belgium       |   11372068|        470|NA                    |NA            |NA      |NA                  |NA       |       NA|         NA|
|BE       |inc hosp        |2021-09-11      |Belgium       |   11372068|         NA|EuroCOVIDhub-ensemble |2021-09-06    |1       |wk                  |point    |       NA|        474|
|BE       |inc hosp        |2021-09-11      |Belgium       |   11372068|         NA|EuroCOVIDhub-ensemble |2021-09-06    |1       |wk                  |quantile |    0.010|        270|
|BE       |inc hosp        |2021-09-11      |Belgium       |   11372068|         NA|EuroCOVIDhub-ensemble |2021-09-06    |1       |wk                  |quantile |    0.025|        304|
|BE       |inc hosp        |2021-09-11      |Belgium       |   11372068|         NA|EuroCOVIDhub-ensemble |2021-09-06    |1       |wk                  |quantile |    0.050|        328|

How can we solve this? :thinking:

Bisaloo:
This is probably what's causing the issue in the visualization app as well.

kathsherratt:
Thanks for looking at this Hugo :) Is the hospitalisation data currently coming from `covidHubUtils::load_truth()`? In the original source ECDC use a Monday-Sunday week, but in either covidHubUtils or our download script I think we load it as a Sunday-Saturday week, in order to match case/death data. This is off the top of my head but may be worth exploring as the raw ECDC data are daily so we really should be able to get the weeks to align.

Bisaloo:
Ah, based on previous conversations, I thought it was a feature and we had to find some kind of interpolation solution. But good to know it's likely a bug :+1:. I'll look into it.

Bisaloo:
Ah my bad, I got confused over nothing. All is fine and as soon as we have truth data for the current week, evaluation reports should work as expected. I'm going to wait until next week to double check and then request a review.

Bisaloo:
Yes, everything works fine now. I'm putting this on hold until we have one more model with hospitalisations, as agreed during the call with the ECDC. But it's ready for review and to be merged otherwise.

Screen Shot 2021-09-17 at 09 12 31-fullpage

Bisaloo:
It's currently failing for Karlen-pypm because the ensemble has not yet been built for this week. I need to figure out how to gracefully handle this situation and make sure it doesn't interrupt the whole process.

PR #792, opened 2021-08-16
Upload 2021-08-15 forecast for Karlen-pypm

deankarlen:


kathsherratt:
Hi @deankarlen - thanks for submitting!

It looks like this PR has a couple of extra stray files. Please could you remove these, before I can merge this in?

As well as the forecast in your team folder, I can see two files in the project root. These are a duplicate `2021-08-15-Karlen-pypm.csv` forecast, and a `testfile.txt`. I guess these are left overs, which just need to be removed. 

Then this PR should only include the `data-processed/Karlen-pypm/2021-08-15-Karlen-pypm.csv` file, and I can merge it directly.

Thanks again,
Kath

deankarlen:
Hi Kath,

  Sorry for the delay in responding… yes, it appears I was having github issues when submitting last weekend. Github stay locked in a “processing files” mode… only workaround was to cut and paste the text of the .csv file. But in so doing it looks like some tests were put into the wrong folders.

  I think I cleaned that mess up, but I guess it is too late for inclusion in the ensemble.

Dean

kathsherratt:
Hi Dean, Thank you for sorting this one! I will re-run the ensemble so it gets included as obviously it should have been.
Kath

PR #801, opened 2021-08-16
Convert viz/ scripts to R and make them more general

Bisaloo:


Bisaloo:
I didn't convert `prepare_forecast_data` yet but I'd appreciate if we can already merge this. There is no point in keeping it unused here and it will prevent spurious diffs because of different locales.

sbfnk (approved):
Looks good to me.

kathsherratt (approved):
Looks good!

PR #803, opened 2021-08-16
Fix warning logical of length > 1

Bisaloo:
If `n_loc` has multiple values, this will throw a warning. Additionally, it there are multiple values but 1 is one of them (e.g. 1, 2, 3), the `n_loc` column will be removed.

This PR implements what I think was the intended goal.


sbfnk (approved):
Yep, good catch. Whilst we're there, should this be `1L`?

PR #816, opened 2021-08-19
Fix truth action

Bisaloo:
See https://github.com/epiforecasts/covid19-forecast-hub-europe/actions/runs/1147068578.


kathsherratt (approved):
👍 good catch!

PR #851, opened 2021-08-30
30 agosto

miraanto:


sbfnk:
Hi, I'm sorry you're having problems with the submission. For information on the problem you can click on "Checks" above and you can see the (admittedly not very user-friendly) error message.

In this case, the problem is that your `target_end_date` is not a Saturday. While we accept forecasts submitted on any weekday, the forecast weeks are still defined as Sunday-Saturday, so the 1 week ahead forecasts made on 29 August need to end on 4 September.

miraanto:
Thanks, yes, I just realized this is the problem and I am uploading the correct file .- sorry for waisting your time, I should have checked before
Best

sbfnk:
No problem! Glad all is working fine now.

PR #893, opened 2021-09-06
Predictions 2021-09-06

BIOCOM-SC:


kathsherratt:
Hi @BIOCOM-SC - thanks for submitting! 

As well as this week's forecast, it looks like you are also adding forecasts from previous weeks (23 and 30 August). Usually, we only include forecasts that are submitted week-by-week as the focus of the hub is on real-time forecasting. 

We are happy to include historic forecasts, we just need to check they were produced at the time (before the forecast overlaps with observed data). Would you have a dated version history of the forecast files from 23 and 30 August that shows when they were produced? (For example, this could be a Github commit history, or a screenshot, e.g. in Windows: file manager > right-click on the file > Properties > Details.) 

If it's not possible to see the version history, please could you let me know any other information about these forecasts?

Best wishes,
Kath

BIOCOM-SC:
> Hi @BIOCOM-SC - thanks for submitting!
> 
> As well as this week's forecast, it looks like you are also adding forecasts from previous weeks (23 and 30 August). Usually, we only include forecasts that are submitted week-by-week as the focus of the hub is on real-time forecasting.
> 
> We are happy to include historic forecasts, we just need to check they were produced at the time (before the forecast overlaps with observed data). Would you have a dated version history of the forecast files from 23 and 30 August that shows when they were produced? (For example, this could be a Github commit history, or a screenshot, e.g. in Windows: file manager > right-click on the file > Properties > Details.)
> 
> If it's not possible to see the version history, please could you let me know any other information about these forecasts?
> 
> Best wishes,
> Kath

Dear Kath,

I'm sorry, I did not see your message until now. During second half of August, I was on holidays. A student was supposed to generate the preditions and upload them. She did it but she did not pèrform the Pull request. I realized it last week, when I made the pull request and I saw that there were several old predictions being pulled. 
So... the addition of old predictions was not voluntary, but a kind of "accident". You can just ignore them or incorporate them in the site, we don't mind because it is, of course, our fault. I attach a screenshot of our GitHub, where you can see the dates when the files were uploaded.

Please, let me know if you need any further clarification. Best regards,

Clara (clara.prats@upc.edu)

image

kathsherratt:
Hi Clara,
Thanks for getting in touch! That is absolutely fine and no problem at all, I will include all the forecasts. 
Best wishes,
Kath

BIOCOM-SC:
Thanks a lot, Kath!

Cheers,

Clara

PR #909, opened 2021-09-13
Forecasts

pmarcosa:
First forecasts + metadata


kathsherratt:
Hi @pmarcosa - thanks for contributing! It's great to have a new forecast.

I noticed that your team name is UNED, the same as an existing team - [`UNED-PreCoV2`](https://github.com/epiforecasts/covid19-forecast-hub-europe/tree/main/data-processed/UNED-PreCoV2). 

It's no problem for teams to submit multiple forecast models. However each team can only have a single model that is designated "primary" in the metadata `team_model_designation` field. Other models must have the `team_model_designation` as "proposed", "secondary", or "other".

Please could you clarify if this is an additional model from the same team, and if so which model is to be considered "primary"?

It would also help here if you could add any extra information about your forecast in the `methods` field, although this doesn't need to be too long.

Thanks very much again!
Kath

pmarcosa:
Hi @kathsherratt, 
I'm an UNED student working on my final project for a masters in data science, that's why I chose "UNED" as my team, but by no means my model should be used as the primary. So I guess "other" should be the correct designation for my model or I can even change the name of the team if needed.
Regarding the extra information, I'm still working offline on the model and so far I just wanted to check if the files would pass the validations (which they did). Apologies if this is not the right way to proceed. I'll add the details of the forecast once it's fully worked out!
An additional question I have is regarding the evaluations. I'd like to evaluate my model offline to see how it compares to existing models in the ensemble before submitting forecasts. My understanding is that before doing that, I have to submit 4 weeks of forecasts. Is that correct? Is it enough if I submit the forecasts to my branch only or does it need to go through the pull request process as well?
Please do let me know if you are not the right person to answer these questions and what's the right channel to post them in that case.

Thanks for your help!
Pablo.

kathsherratt:
Hi @pmarcosa - I'm so sorry for the delay replying. Hope this helps.

> So I guess "other" should be the correct designation for my model or I can even change the name of the team if needed.

Re team name, I guess it depends if you have developed this model together with the same people that submit the  UNED-PreCoV2 model. If you have developed it independently, it might be best to use a different team name for your own model.

> Regarding the extra information, I'm still working offline on the model and so far I just wanted to check if the files would pass the validations (which they did). 

That's good to know, thanks. Is that still the case, or would you consider the two forecasts in this PR ready for submission? If you think they are ready I can merge them in. Otherwise, it's no problem to keep working on the model, but if you don't think they are ready for submission please could you close this pull request, and open a new PR when your forecasts are ready? As a reminder, you can still check that the forecasts pass validations locally without opening a new PR: see [here for instructions](https://github.com/epiforecasts/covid19-forecast-hub-europe/wiki/Check-submission).

> An additional question I have is regarding the evaluations. I'd like to evaluate my model offline to see how it compares to existing models in the ensemble before submitting forecasts. My understanding is that before doing that, I have to submit 4 weeks of forecasts. Is that correct? Is it enough if I submit the forecasts to my branch only or does it need to go through the pull request process as well?

Yes, we only evaluate models that have been submitted and merged into the hub, and report the evaluations on the [website](https://covid19forecasthub.eu/reports.html), as you may have seen. 

To do this we use the [`scoringutils`](https://github.com/epiforecasts/scoringutils) package. However we do have a lot of code that goes into the evaluation, so I suggest three options depending on how detailed you would like to replicate the formal hub evaluation process:

1) The outline of how we create the reports each week lives [here](https://github.com/epiforecasts/covid19-forecast-hub-europe/blob/main/code/reports/evaluation/evaluation-report.Rmd). 

2) I've put together a (slightly) simpler version which can be run locally to create a table of forecast scores for all forecasts you have saved in the `data-processed` folder. The [gist is here](https://gist.github.com/kathsherratt/bcd6f2718a77e73a8197dfb453e416a6). 

3) If you know which evaluation metric you want to look at, that code might still be a bit over-complicated so you may be best using the  [`scoringutils`](https://github.com/epiforecasts/scoringutils) package directly to evaluate your forecasts.

Let me know if any of that helps, and if you would like me to merge the forecasts in this PR (with a change in team name/designation).

All the best,
Kath

PR #922, opened 2021-09-13
Filter out forecasts that go further than the specified horizons 

Bisaloo:


sbfnk (changes requested):
is `rmdchunks/get-config-vars.Rmd` missing?

sbfnk (approved):

PR #983, opened 2021-09-28
split scores

sbfnk:
Updates to work with https://github.com/epiforecasts/EuroForecastHub/pull/5

### DIFF
--- a/code/evaluation/score_models.r
+++ b/code/evaluation/score_models.r
@@ -85,5 +85,3 @@ for (chr_report_date in as.character(report_dates)) {
 
   write_csv(table, eval_filename)
 }
-
-write_csv(scores, here::here("evaluation", "scores.csv"))
--- a/code/evaluation/score_models.r
+++ b/code/evaluation/score_models.r
@@ -52,6 +52,15 @@ data <- scoringutils::merge_pred_and_obs(forecasts, truth,
 latest_date <- today()
 wday(latest_date) <- get_hub_config("forecast_week_day")
 
+message("Scoring all forecasts.")
+
+scores <- score_forecasts(
+  forecasts = data,
+  quantiles = get_hub_config("forecast_type")$quantiles
+)
+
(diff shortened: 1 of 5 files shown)


Bisaloo:
We still need to update the GitHub actions:

- create a new action to run `score_models.r`
- remove the csv uploading step from the current action

It could also be a single action with two jobs (1. `score_models.r`, 2. `compile_evaluation_report.r`). The second would be dependent on the success of the first one.

sbfnk:
> It could also be a single action with two jobs (1. `score_models.r`, 2. `compile_evaluation_report.r`). The second would be dependent on the success of the first one.

Probably fine to be a single action - I can't really imagine a situation where we'd want to, e.g., re-run one without the other.

code/evaluation/score_models.r
```diff
@@ -63,15 +63,20 @@ report_dates <- seq(start_date, latest_date, by = "week")
...
   report_date <- as.Date(chr_report_date)
-  filename <-
+  eval_filename <-
     here::here("evaluation", paste0("evaluation-", report_date, ".csv"))
 
-  table <- score_models(
-    data,
-    report_date,
-    restrict_weeks = 4,
+  scores <- score_forecasts(
```
Bisaloo, line 69: It doesn't seem super efficient to include this in the loop. In the case where we have to re-run all the evaluations (which is precisely what we will have to do once this is merged), we are going to repeat the same operations many times since scores(t+1) are just a superset of scores(t).

sbfnk: Yes, good point - in fact I already did this in [this experimental branch](https://github.com/epiforecasts/covid19-forecast-hub-europe/blob/a2c232f796a7b7baff21872c9f9d89553b090e1f/code/evaluation/score_models.r#L60). I'll copy over the change from there.

Bisaloo (approved):

PR #986, opened 2021-09-30
Upload files for angular app to Azure

Bisaloo:


viz/prepare_forecast_data.py
```diff
@@ -29,7 +29,8 @@ def get_relevant_dates(dates):
 models_to_include = [k for k,v in metadata.items() if (v['team_model_designation'] != 'other')]
 
 VALID_TARGETS = [f"{_} wk ahead inc death" for _ in range(1, 5)] + \
-                [f"{_} wk ahead inc case" for _ in range(1, 5)]
+                [f"{_} wk ahead inc case" for _ in range(1, 5)] + \
```
sbfnk, line 32: I think this should be `range(1, 4)` in both instances.

sbfnk (approved):

viz/prepare_forecast_data.py
Bisaloo: It's a bit counter-intuitive but:

```py
VALID_TARGETS = [f"{_} wk ahead inc death" for _ in range(1, 5)] + \
                [f"{_} wk ahead inc case" for _ in range(1, 5)] + \
                [f"{_} wk ahead inc hosp" for _ in range(1, 5)]
VALID_TARGETS
```

> ['1 wk ahead inc death', '2 wk ahead inc death', '3 wk ahead inc death', '4 wk ahead inc death', '1 wk ahead inc case', '2 wk ahead inc case', '3 wk ahead inc case', '4 wk ahead inc case', '1 wk ahead inc hosp', '2 wk ahead inc hosp', '3 wk ahead inc hosp', '4 wk ahead inc hosp']

[The stop argument is not included in the output](https://www.w3schools.com/python/ref_func_range.asp)

sbfnk: Ah well it‘s python so it doesn‘t have to make sense 😉

PR #987, opened 2021-09-30
Re-run evaluations

Bisaloo:
follow up of https://github.com/epiforecasts/covid19-forecast-hub-europe/pull/983


sbfnk (approved):
Great. We should also re-run the report (can do from the github action).

Bisaloo (commented):
I need to fix sorting so we don't have the usual problem of spurious diffs

Bisaloo (commented):
> I need to fix sorting so we don't have the usual problem of spurious diffs

I *think* the order comes directly from the data we feed to `eval_forecasts()`, which means it is already fully reproducible because it has been fixed in `covidHubUtils::load_forecasts()`.

PR #990, opened 2021-10-01
remove LANL action

sbfnk:
LANL has stopped producing forecasts, see https://covid-19.bsvgateway.org/


Bisaloo:
Should we remove https://github.com/epiforecasts/covid19-forecast-hub-europe/blob/main/code/auto_download/auto-download-lanl-covid19.r at the same time? It will always be available in the history anyways

sbfnk:
Ah, yes, please do.

PR #997, opened 2021-10-04
calculate recent scores in addition to scores with all history

sbfnk:
Also renames `score-forecasts.Rmd` to `load-scores.Rmd`

Results in three rows of tabs like this
image

(I've checked that the scores haven't changed, but the added column means git can't just diff)


sbfnk:
Let's go ahead.

PR #1015, opened 2021-10-05
Add LANL custom license

Bisaloo:


sbfnk:
According to the [wiki](https://github.com/epiforecasts/covid19-forecast-hub-europe/wiki/Licensing) if a LICENSE.txt is provided the `license:` field in the metadata should be set to `LICENSE.txt`. 

That said, I'm not sure this makes sense - surely if license is `other` we need a LICENSE.txt, and also LICENSE.txt is not in the [list of accepted licenses](https://github.com/reichlab/covid19-forecast-hub/blob/master/code/validation/accepted-licenses.csv). Shall we update the wiki to say that if a LICENSE.txt is provided the license should be set to `other`?

Bisaloo:
> Shall we update the wiki to say that if a LICENSE.txt is provided the license should be set to other?

I'd prefer the opposite: if the license is set to other, then a `LICENSE.txt` file should be provided.

sbfnk:
> > Shall we update the wiki to say that if a LICENSE.txt is provided the license should be set to other?
> 
> I'd prefer the opposite: if the license is set to other, then a `LICENSE.txt` file should be provided.

Yes, that way around it makes more sense.

sbfnk (approved):

PR #1016, opened 2021-10-05
Metadata emails

kathsherratt:
A small PR to help process email addresses in metadata
- Corrects formatting of a few email addresses in metadata
- Adds code to get all metadata emails and cross-check against those in the google group


Bisaloo (approved):
LGTM. Two minor comments:

- should we add epiforecasts email addresses for the baseline?
- the next time we want to build a hub, it would probably be simpler to have a separate field for email addresse(s) rather than having them in the authors (`model_contributors`) field

sbfnk:
Both good points - can add your email to the baseline given that it was you (I think) who implemented it for us?

PR #1025, opened 2021-10-07
Update metadata-CovidMetrics-epiBATS.txt

arodloff:
Typo in e-mail address


arodloff:
> Hi @arodloff 👋, thanks for the pull request.
> 
> I had to make some changes to the metadata file earlier this week because it could not be read by our tools in its current form. So the changes you propose here cannot be integrated directly because of a version clash with our latest version (git conflict). I think the simplest solution is for me to copy this change manually on the latest version. Unless you disagree with this course of action, there is no action necessary on your end for this.
> 
> However, while you are here, could you also please have a look at your `methods` section? We have a 200 characters limit and yours is currently 245 characters long. This is necessary since this is displayed on the website as a tooltip and long texts take all the space. Could you send us a shorter version please? You are free to expand in the `methods_long` section.
> 
> Thanks,
> 
> Hugo

Hi @Bisaloo,
thank you for your reply and implementing the change manually. You can also change the methods text to "Forecasts are based on TBATS - models (DeLivera, Hyndman and Snyder (2011)) and are updated daily for each German state. " (unless there is a better way to implement this change). 

Arne

PR #1031, opened 2021-10-11
Re-naming KITmetricslab-bivariate_branching, shortening model description.

jbracher:


data-processed/KITmetricslab-bivar_branching/metadata-KITmetricslab-bivar_branching.txt
```diff
@@ -1,21 +1,19 @@
 team_name: KITmetricslab
-model_name: bivariate_branching
```
Bisaloo, line 2: You can actually keep this line as it was if you wish. The most important is to edit the line below: `model_abbr`. It is the one with the character limit

jbracher:
Sorry, my bad

PR #1056, opened 2021-10-13
2021-10-13 ECDC hosp_model 

RokGrah:


kathsherratt:
Hi Rok, thanks for submitting, great to have the forecasts!

Sorry you have had trouble here. The issue is that we don't hold data for all countries - as we've discussed, we take the daily data from ECDC and filter to web-scraped sources only. That means that the validation will fail if there are forecasts in locations that we don't have data for. 

You can see which locations have data in this csv: https://github.com/epiforecasts/covid19-forecast-hub-europe/blob/main/data-locations/locations_eu.csv
where the `inc_hosp` field has a non-NA value indicating where the data are sourced from ("ECDC").

I think there are only 11 countries there with hospitalisation data - if you remove forecasts for locations elsewhere, validation checks should pass fine.

Let me know if you have any other trouble.
Kath

RokGrah:
Hi Kath,

one way to solve the error mentioned above is to just set the forecast date to Monday, 2 days ago. We did not use any data from after Monday so it would be entirely wrong. Can you confirm that this would be ok from your side?

Thanks,
Rok

PR #1063, opened 2021-10-13
Statgroup19 first submission only metadata

minmar94:
Create folder with metadata only


kathsherratt:
Hi @minmar94 - thanks for joining the forecast hub!

This metadata looks good to me. If this is the only model from your team, you could set your `team_model_designation` to "primary". (We use the "proposed" designation when looking at multiple models from the same team.)

Looking forward to your forecasts.

Best wishes,
Kath

PR #1137, opened 2021-10-29
Reduce report size and prepare them to get out of iframe

Bisaloo:


code/reports/country/country-report.Rmd
```diff
@@ -5,13 +5,13 @@ params:
   location_name: "Overall"
   plot_weeks: 4
   data_weeks: 10
+  config: "https://raw.githubusercontent.com/epiforecasts/covid19-forecast-hub-europe/main/forecasthub.yml"
```
sbfnk, line 8: Should this be a local path? Don't feel strongly but it seems that if e.g. I wanted to change the config in a branch I'm working on then this should filter through to these.

sbfnk (changes requested):
Great stuff!

code/reports/country/country-report.Rmd
Bisaloo: It's just the default value. You can change it for a local path when you are developing. In production, we need a URL because the reports will live in the `-website` repo. A local path cannot work with GitHub actions.

If you wish, I can keep a local path as default and just put the URL in the `compile-...-reports.r` script.

code/reports/rmdchunks/load-data.Rmd
```diff
@@ -28,7 +28,7 @@ raw_truth <- load_truth(
 )
 
 # get anomalies
-anomalies <- read_csv(here("data-truth", "anomalies", "anomalies.csv"))
+anomalies <- read_csv(paste0(hub_repo, "/raw/main/data-truth/anomalies/anomalies.csv"))
```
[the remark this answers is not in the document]
Bisaloo: This works on windows as well since it's not an actual path but a URL (`hub_repo = "https://github.com/epiforecasts/covid19-forecast-hub-europe"`). I can still change it. I didn't do it because I was thinking I needed to add URL params but turns out it's not necessary.

code/reports/country/country-report.Rmd
sbfnk: No, this makes sense.

code/reports/rmdchunks/load-data.Rmd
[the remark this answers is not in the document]
sbfnk: Of course. My bad.

sbfnk (approved):

Issue #941, opened 2021-09-20
Calculation of hospitalisation data


sbfnk:
At the moment our hospitalisation data is subject to a fair amount of retrospective revision.

In order to be consistent with the JHU data we're using what we should probably do instead is
1) compile a daily data set of cumulative number of hospitalisations per country
2) every week, take the Sunday-Sunday (or Saturday-Saturday) difference as weekly new number of hospitalisations.

Bisaloo:
I have a script doing this for the data we recorded until now. But the results don't look so great. Especially because we have some gaps in the data (weekends?)

``` r
library(gh)
library(purrr)

get_all_versions <- function(file, gh_repo) {

  gh(
    "/repos/{gh_repo}/commits?path={file}",
    gh_repo = gh_repo,
    file = file
  ) %>%
    map_chr("sha") %>%
    map_chr(~ glue::glue(
      "https://raw.githubusercontent.com/{gh_repo}/{sha}/{file}",
      gh_repo = gh_repo,
      sha = .x,
      file = file
    )) %>%
    map_chr(utils::URLencode) %>%
    map(readr::read_csv, show_col_types = FALSE, progress = FALSE)

}

all_hospdata_versions <- get_all_versions(
  "data-truth/ECDC/truth_ECDC-Incident Hospitalizations.csv",
  "epiforecasts/covid19-forecast-hub-europe"
)

library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
adjusted_hospdata <- all_hospdata_versions %>%
  map(group_by, location) %>%
  map(arrange, date) %>%
  map(mutate, cum_value = cumsum(value), .keep = "unused") %>%
  map(filter, date == max(date)) %>%
  bind_rows() %>%
  distinct() %>%
  arrange(location, date) %>%
  mutate(value = cum_value - lag(cum_value))

adjusted_hospdata
#> # A tibble: 225 × 5
#> # Groups:   location [11]
#>    date       location location_name cum_value value
#>    <date>     <chr>    <chr>             <dbl> <dbl>
#>  1 2021-08-27 BE       Belgium           77587    NA
#>  2 2021-08-30 BE       Belgium           77758   171
#>  3 2021-08-31 BE       Belgium           77842    84
[...]

PR #1169, opened 2021-11-04
Switch hospitalisation data sources

kathsherratt:
Overview below, plus detail in a [README](https://github.com/epiforecasts/covid19-forecast-hub-europe/blob/hosp-compare/code/auto_download/hospitalisations/README.md). This handles sources to create a single hospitalisations dataset (expanding data to 21 countries 🎉 ).

Sources are:
- ECDC scraped daily
- ECDC official weekly
- non-EU scraped daily

Sources are selected automatically using code from @sbfnk:
- selection done in: `code/auto_download/hospitalisations/check-sources/select_hosp_sources.r`
- csv output of selected sources / truncation: `code/auto_download/hospitalisations/check-sources/sources.csv`)

Source selection was run once and used as a fixed output for now - but maybe we could check this again / on some regular schedule (as I think ECDC mentioned they were adding new countries at some point). And/or switch to report dates as well (#971, #941).

---

A new data update pipeline runs in `code/auto_download/hospitalisations-download.R`. This does the following:

1. Each source is downloaded and cleaned. 
  - Each of the three sources is a separate script (`hospitalisations/get-[...].R`)
  -  All dates are kept as they are in raw data (ie daily or Mon-Sun week used by ecdc). 
  - Each saves into a separate csv in `./data-truth/ECDC/raw/`

2. Sources are combined and selected in `hospitalisations/save-selected-sources.R`
  - Daily data are aggregated to weeks (all Mon-Sun)
  - All data are combined
  - One source is kept for each country (using `hospitalisations/check-sources/sources.csv`)
  - The end date for the week period of all data is shifted back -1 to Saturday, to match case/death data MMWR weeks.
  - This new dataset is saved as `data-truth/ECDC/truth_ECDC-Incident Hospitalizations.csv`.
 
3. Validation needs to know which countries have hospitalisation data so are valid targets for `inc hosp` forecasts.
  - Currently this uses `data-locations.csv`. This csv should therefore be updated whenever countries with new hosp data are added/removed. 
  - Code to create/update `data-locations.csv` is now saved in `create-data-locations.R`. This uses `data-truth/ECDC/truth_ECDC-Incident Hospitalizations.csv` to identify which locations have hospitalisation data
  - `create-data-locations.R` is included in every run of the hospitalisation data pipeline (`hospitalisations-download.R`).

---
Wider code impact:

- The daily github action (`./.github/workflows/ECDC.yml`) is updated to use `hospitalisations-download.R`.
- Validation should cope fine as it will just see a longer list of valid locations for the `inc hosp` target in `data-locations.csv`.
- This all still works with `covidHubUtils::load_data()` to load the final combined/selected dataset (`data-truth/ECDC/truth_ECDC-Incident Hospitalizations.csv`).
```
covidHubUtils::load_truth(truth_source = "ECDC",
                          target_variable = "inc hosp",
                          local_repo_path = here(),
                          hub = "ECDC")
```


code/auto_download/hospitalisations-download.R
```diff
@@ -0,0 +1,16 @@
+# Download and save all hospitalisations data
+#  plus save a "truth" file with only the selected source for each country
+
+library(here)
+script_dir <- here("code", "auto_download", "hospitalisations")
+
+# Download and save minimally processed data files
+source(here(script_dir, "get-ecdc-official.R"))
```
sbfnk, line 8: Should this (and the next few lines) be `file.path` instead of `here` as line 5 has already added the necessary folder structure? I presume you've tested it and it works in which case I've learned something.

code/auto_download/hospitalisations/README.md
```diff
@@ -0,0 +1,70 @@
...
+    -   Origin: Mix of public data scraped from websites (typically
+        government data) by the ECDC, and data sent to ECDC directly
+        through the TESSy European data sharing agreement. All data are
+        processed and vetted by ECDC
+    -   Released: weekly on Thursdays. Each data point covers the
+        previous Monday-Sunday period. Data are incidence per 100,000
+        population.
+    -   Download with:
+        [`get-ecdc-official.R`](./code/auto_download/hospitalisations/get-ecdc-official.R)
+-   ECDC scraped:
```
sbfnk, line 37: As it's not an official ECDC dataset perhaps we should remove ECDC here for clarify; could call it "EU scraped" perhaps?

code/auto_download/hospitalisations/README.md
```diff
@@ -0,0 +1,70 @@
...
+    [data-truth/ECDC/raw/scraped.csv](./data-truth/ECDC/raw/scraped.csv)
+    -   Source: Confidential data released by ECDC to the European
```
sbfnk, line 39: It's not really confidential (given that it's web-scraped it's the opposite). Perhaps say "Additional data obtained from public sources and provided by the ECDC to the European Forecast Hub"?

code/auto_download/hospitalisations/get-non-eu.R
```diff
@@ -0,0 +1,37 @@
...
+
+# Format ------------------------------------------------------------------
+non_eu <- bind_rows(uk, ch) %>%
+  select(date, location_name, hosp_new) %>%
+  left_join(covidHubUtils::hub_locations_ecdc %>%
+              select(-population),
+            by = "location_name") %>%
+  select(location_name, location, date, value = hosp_new) %>%
+  mutate(source = "Public",
+         type = "National")
```
sbfnk, line 34: Should we call this "Scraped (hub)" or "Scraped (non-EU)" instead of "National" (as all the data are national)?

code/auto_download/hospitalisations/get-ecdc-official.R
```diff
@@ -0,0 +1,26 @@
+# Get official weekly ECDC hospitalisation data
+library(dplyr)
+library(covidHubUtils)
```
sbfnk, line 3: I think this needs `library(readr)` too for `read_csv` below

sbfnk (approved):
This is really great, thank you! Very happy for this to be merged.

One remaining issue is that the truncated forecasts created issues in the [visualisations](https://covid19forecasthub.eu/visualisation.html?target=hosp&location=BE&models=ECDC-hosp_model&models=EuroCOVIDhub-ensemble&models=Karlen-pypm&models=MUNI-ARIMA&models=USyd-OneModelMan&models=epiforecasts-arimareg&models=epiforecasts-caseconv&models=epiforecasts-tsensemble) but we can deal with this separately.

code/auto_download/hospitalisations-download.R
kathsherratt: I think `here` handles this with its own magic (so I just use `here` everywhere). In this case
```
> file.path(script_dir, "get-ecdc-official.R") == here(script_dir, "get-ecdc-official.R")
[1] TRUE
```

PR #1193, opened 2021-11-09
Upload 2021-11-07 Forecast for Karlen-pypm

deankarlen:
Includes new hospitalization forecasts. Note PL, PT, CH hospitalization levels are much too low: no forecast included.


kathsherratt:
Hi @deankarlen - thanks for flagging the data for Poland, Portugal, and Switzerland. The updates on these (some you may already have seen) are:

- Poland - we have now excluded this country as a forecast target after finding not all regions are included in the national total.
- Switzerland - we found a problem with data processing which we have now fixed, so numbers now look more as expected. I'm sorry about the problem this week.
- Portugal - I can't triangulate the low numbers against any other source for admissions, as the ECDC source we use is the only one available. 
   - After a quick look at other data I think these numbers are reasonable (at something like 5-10 admissions per day recently). I can see similar-ish changes in occupancy (around +/- 10-20 patients daily, [available here](https://github.com/dssg-pt/covid19pt-data/blob/master/data.csv)) and numbers of [deaths](https://github.com/epiforecasts/covid19-forecast-hub-europe/blob/main/data-truth/JHU/truth_JHU-Incident%20Deaths.csv) (~ 5 daily) looking over October. Both are also very low relative to cases, bearing in mind Portugal has one of the highest vaccination rates in Europe. 
   - Tagging @sbfnk for a second opinion though, and let me know if you think the numbers are still off, no problem to check again with the ECDC.

PR #1199, opened 2021-11-09
2021-11-08 ECDC hosp model

RokGrah:


kathsherratt:
Hi @RokGrah - it looks like your forecast is saved in the project root directory, rather than the `ECDC-hosp_model` team folder within `data-processed` ([here](https://github.com/epiforecasts/covid19-forecast-hub-europe/tree/main/data-processed/ECDC-hosp_model)). Please could you move the file to the correct folder? Only then the validations will run and we can merge this in. 
Thanks!
Kath

PR #1198, opened 2021-11-09
2021-11-07 epiforecasts-arimareg, epiforecasts-caseconv, epiforecasts-tsensemble

sophiemeakin:

PR #1200, opened 2021-11-09
Sophie Meakin's automated submission

sbfnk:


Bisaloo:
Redundant with #1198 but it should work next time :grimacing:. Sorry about this.

Bisaloo:
Actually, this might be better than #1198 because https://github.com/epiforecasts/covid19-forecast-hub-europe/commit/2809ede0aa2692c9409ee1ded0b7dd9c97eceaa0 was added in the meantime and this causes a huge difference in hospitalization forecasts for Switzerland.

sophiemeakin (commented):
All looks good 👍🏻

kathsherratt:
> Actually, this might be better than #1198 because [2809ede](https://github.com/epiforecasts/covid19-forecast-hub-europe/commit/2809ede0aa2692c9409ee1ded0b7dd9c97eceaa0) was added in the meantime and this causes a huge difference in hospitalization forecasts for Switzerland.

We will exclude Switzerland this week - #1197 - so shouldn't be a problem either way, but all fine anyway.

PR #1204, opened 2021-11-10
Use svg instead of jpg for hosp truth data plot

Bisaloo:
This results in:

- smaller file size (100KB vs 500KB)
- clearer details in the plot


kathsherratt:
Great and thank you very much for adding the alt text too.

kathsherratt (approved):

PR #1232, opened 2021-11-15
2021-11-15 MUNI-ARIMA, -VAR, -LaggedRegARIMA (new model)

kraus-stat:
This PR contains the first submission of a new model, MUNI-LaggedRegARIMA. Since it is related (although not near-identical) to MUNI-VAR, we designate it as "other" initially. For now our goal is to see the evaluation scores. Based on that, we may change the designation and reorganize our models.


kathsherratt:
That's great, thank you for the update @kraus-stat!

PR #1245, opened 2021-11-21
Forecast submission Statgroup19

minmar94:


kathsherratt:
Hi Marco @minmar94 , thank you for submitting! 
In the `spatialrichards` model, it looks like there is an error with the `target_end_date` field. For example there is a forecast for 2021-11-20, and other forecasts are one week out (e.g. 2021-11-27 is given as a 2-week ahead forecast, this should be 1 week ahead). Please could you check this? 
Best wishes,
Kath

minmar94:
Dear Kath,

thank you for pointing out the error, it was my fault. I just submitted the
correct version of the forecast.
Best regards,
Marco Mingione


Il giorno lun 22 nov 2021 alle ore 11:52 Kath Sherratt <
***@***.***> ha scritto:

PR #1273, opened 2021-11-23
2021-11-21 epiforecasts-tsensemble, epiforecasts-arimareg, epiforecasts-caseconv

sophiemeakin:
Updated forecasts to deal with truncated data correctly. Forecasts are made for locations with no truncation (DK, EE, FR, SI) or one week of truncation (BE, GB, NO). Where data is truncated for `x` weeks, the models use a `4+x`-week horizon. Can easily extend this to other locations where there is longer truncation, will discuss separately.


sophiemeakin:
Looking into the error now - probably just an issue that I didn't catch when updating to use the truncated data.

kathsherratt:
Thanks @sophiemeakin (and sorry for the trouble earlier)! Looks like some of your forecast dates are from last week (`2021-11-14`), please could you check?

PR #1286, opened 2021-11-29
forecast hosp

FohmAnalys:


FohmAnalys:
Hi @kathsherratt ,

Unsure what to do about `location invalid for target 'inc_hosp': 'SE'`

kathsherratt:
Hi @FohmAnalys - thank you for contributing.

Unfortunately at the moment we don't forecast hospitalisations for Sweden. This is because we don't have access to a source of incident general hospital admissions data in Sweden, against which to evaluate forecasts. We use the [ECDC data](https://www.ecdc.europa.eu/en/publications-data/download-data-hospital-and-icu-admission-rates-and-current-occupancy-covid-19), and this does not include general admissions for Sweden (only occupancy and incident ICU).

There is more information about the countries and data we include for hospitalisations available [here](https://github.com/epiforecasts/covid19-forecast-hub-europe/tree/main/data-truth#hospitalisations) and [here](https://github.com/epiforecasts/covid19-forecast-hub-europe/tree/main/code/auto_download/hospitalisations#readme).

Best,
Kath

kathsherratt:
Hi @FohmAnalys - the hub team discussed this further, and we can merge this forecast for inclusion in the hub. Thank you again for contributing!

As above, we don't have data for Sweden at the moment, which means this won't show up in the visualisation or evaluation of forecasts on the website. But if Swedish data becomes available through the ECDC in future, we will be able to use these forecasts for both those purposes then. 

You are also very welcome to join the mailing list for the hub, which you can request [here]( https://groups.google.com/u/1/g/euro-covid19-forecast-hub). We use this mailing list for very occasional updates on data, and to send weekly invites to meetings (Wednesdays, 1400 CET), which anyone is welcome to attend.

Kath

Bisaloo:
Hi @FohmAnalys, there is an small issue we overlooked with your submission. Your metadata file has an field `methods_long` without any content. Could you remove this field entirely and submit a new pull request please?

PR #1292, opened 2021-11-29
2021-11-29 epiMOX-SUIHTER

giovanniardenghi:


kathsherratt:
Hi @giovanniardenghi - thank you as always for contributing. 

I noticed you had previously uploaded hospital forecasts, which failed checks in this PR - I'm sorry about that. We recently had to exclude hospital admissions data from Italy as we saw issues with data quality, which caused the validation check on this PR to fail.

However, from this week we do still accept and include forecasts for locations which we don't have data for, including Italy. Please feel free to re-add your hospitalisation forecasts to this PR, or just continue to include them in future.

Unfortunately, we won't be able to run the weekly visualisation or evaluation on hospitalisation forecasts for locations without data. Still, if data quality improves in future and we can re-introduce Italy as a forecast target, these forecasts will then be used for both those purposes.

Thanks again,
Kath

PR #1305, opened 2021-11-30
2021-11-28 epiforecasts-arimareg, epiforecasts-caseconv, epiforecasts-tsensemble

sophiemeakin:
Correct forecast date that caused error in original auto submission (https://github.com/epiforecasts/covid19-forecast-hub-europe/pull/1303)


Bisaloo:
Don't worry about the failing check. It's caused by another model and it will not block this PR.

PR #1306, opened 2021-11-30
2021-11-29 ECDC hosp model

RokGrah:


kathsherratt:
Thanks @RokGrah. Just to note the failing check is an issue on our side so nothing to worry about.

PR #1347, opened 2021-12-13
2021-12-13 - JBUD submission

joebud:


kathsherratt:
Hi @joebud - thanks for submitting. 

Unfortunately it looks like your forecast end dates are out by 1 day. [We use MMWR weeks](https://github.com/epiforecasts/covid19-forecast-hub-europe/wiki/Targets-and-horizons#date-format), which are Sunday - Saturday, so all `target_end_date` values should be Saturday (e.g. 2021-12-19 should be replaced with 2021-12-18). The checks should pass when this is updated.

Kath

PR #1446, opened 2022-01-05
Add MOCOS for 2022-01-03

dwolffram:
This submission to the German/Polish Hub was delayed, that's why I'm forwarding it separately.


dwolffram:
Sorry, forgot to reformat.

Issue #601, opened 2021-07-01
Use `yml` extension for metadata files


Bisaloo:
Currently, metadata files must have the format `metadata-model.txt` but it looks like a YAML file and it's actually read using the `yaml` python library in the `validations` submodule.

Would it make sense to use the `yml` file extension instead?

Benefits:

1. We get syntax highlighting, which might help spotting syntax error. Below is an example where the user adds `:` without enclosing the string in quotes ( `source:` appears in brown):

![image](https://user-images.githubusercontent.com/10783929/124135503-62495780-da73-11eb-83f7-e478e875da66.png)

2. Users are explicitly allowed to use comments with the usual YAML syntax

Bisaloo:
Lines to edit if we want to change this:

https://github.com/epiforecasts/covid19-forecast-hub-europe-validations/blob/31c2cffbc25bea944bdc184260e96aad4ea6cd27/main.py#L25

https://github.com/epiforecasts/covid19-forecast-hub-europe-validations/blob/31c2cffbc25bea944bdc184260e96aad4ea6cd27/main.py#L153

https://github.com/epiforecasts/covid19-forecast-hub-europe-validations/blob/31c2cffbc25bea944bdc184260e96aad4ea6cd27/main.py#L171

https://github.com/epiforecasts/covid19-forecast-hub-europe-validations/blob/31c2cffbc25bea944bdc184260e96aad4ea6cd27/main.py#L176

https://github.com/epiforecasts/covid19-forecast-hub-europe-validations/blob/31c2cffbc25bea944bdc184260e96aad4ea6cd27/codebase/validation_functions/metadata.py#L28

https://github.com/epiforecasts/covid19-forecast-hub-europe-validations/blob/31c2cffbc25bea944bdc184260e96aad4ea6cd27/codebase/validation_functions/metadata.py#L156

https://github.com/epiforecasts/covid19-forecast-hub-europe-validations/blob/31c2cffbc25bea944bdc184260e96aad4ea6cd27/codebase/validation_functions/metadata.py#L170

sbfnk:
It would definitely be an improvement - only downside I can see is that if we rename all the metadata files then if teams work in a fork and never merge `main` (which I think is how teams operate in most cases) then if they ever change their metadata the corresponding PR can't be merged without them explicitly renaming the file. This might lead to some confusion.

That said, it's probably worth the benefit, and we could flag it to the teams so they're aware.

[...]

PR #1475, opened 2022-01-15
Compare infrastructure for a new scenario hub

kathsherratt:
Adding a document to suggest how we can adapt the existing forecast hub repo structure to a new scenario hub - by comparison with the US scenario hub. https://github.com/midas-network/covid19-scenario-modeling-hub

This is only about bare-bones infrastructure, I haven't covered any of the actual scenario-setting work her - just the basic differences in folders, files, validation that would cover the technical side of setting up a new hub compared to what we have in the forecast hub.

Happy to move this markdown somewhere else, just adding here for now.

### DIFF
--- a/docs/scenario-hub-comparison.md
+++ b/docs/scenario-hub-comparison.md
@@ -1,0 +1,81 @@
+# Euro scenario hub set up
+
+This document outlines where the infrastructure for the European COVID-19 Forecast Hub varies from that of the US COVID-19 Scenario hub.
+
+This covers the differences between Github repo structure only, including the [folder structure](#main-branch-folder-structure) and the format for [individual file submissions](#changes-to-team-submissions).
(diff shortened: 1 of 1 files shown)


Bisaloo:
Thanks for this Kath, it is amazingly helpful!

I'm currently trying to create our infrastructure and I have a couple of comments:

- in the forecast files, `scenario_name` and `scenario_id` seem redundant. Should we keep only one of them? If so, which one?
- I believe we should use this opportunity to add a couple of (minor) breaking changes: use the yaml/yml extension in metadata files (#601), clarify which licenses are valid and add them to the validations

kathsherratt:
With many thanks to @Bisaloo, I think all of this text has now moved over to the scenario hub wiki, so closing this PR.

PR #1482, opened 2022-01-17
Add folder and metadata for epiforecasts-weeklygrowth

kathsherratt:
Should mean that #1474 can be updated automatically and checks will work to validate forecasts, without failing because of missing metadata


Bisaloo:
@kathsherratt, is it possible to merge this already? It will then be simpler to look at the other PR.

PR #1509, opened 2022-01-21
Convert yml config file to json

Bisaloo:
to match name and format with US hub

Code:

```r
config <- yaml::read_yaml("forecasthub.yml")
jsonlite::write_json(config, "project-config.json", pretty = TRUE, auto_unbox = TRUE)
```

We still need to agree with the US hub on fields but this is a first step in the direction of compatibility.


Bisaloo:
Must be synced with https://github.com/covid19-forecast-hub-europe/EuroForecastHub/pull/12.

sbfnk (approved):
I have no preference re yaml vs. json - good to be in sync with US hub(s).

PR #1537, opened 2022-01-25
Remove stale hospitalisation data

kathsherratt:
This PR updates the data sourcing process to exclude data that is over 16 days old from the present date. This currently removes Greece.

We could also add a check or flag where there are long gaps between data updates. This would cover Denmark, which updated on the 11 December and then next on 22 January. I'm not sure we would want to remove the location from the data entirely, but we'd presumably want to add just those weeks in Denmark to the list of exclusions (to remove from evaluation any forecasts submitted in that time).


code/auto_download/hospitalisations/save-selected-sources.R
```diff
@@ -57,14 +57,25 @@ non_eu <- read_csv(here(data_dir, "raw", "non-eu.csv")) %>%
...
-  select(location_name, location, date, value, selected_source, origin)
+                 non_eu %>% mutate(selected_source = TRUE))
+
+# Check selected data are fresh (< 2 weeks old) -----------------------
+location_stale <- all %>%
+  filter(selected_source) %>%
+  group_by(location, source, type) %>%
+  summarise(n = n(),
+            max_date = max(date)) %>%
+  mutate(stale = ifelse(max_date < Sys.Date() - 16, TRUE, FALSE)) %>%
```
sbfnk, line 70: can shorten to `mutate(stale = (max_date < Sys.date() - 16)`

code/auto_download/hospitalisations/save-selected-sources.R
```diff
@@ -57,14 +57,25 @@ non_eu <- read_csv(here(data_dir, "raw", "non-eu.csv")) %>%
...
-                   mutate(selected_source = TRUE)) %>%
-  mutate(origin = paste(source, type, sep = "-")) %>%
-  select(location_name, location, date, value, selected_source, origin)
+                 non_eu %>% mutate(selected_source = TRUE))
+
+# Check selected data are fresh (< 2 weeks old) -----------------------
+location_stale <- all %>%
+  filter(selected_source) %>%
+  group_by(location, source, type) %>%
+  summarise(n = n(),
```
sbfnk, line 68: `n = n()` needed?

code/auto_download/hospitalisations/save-selected-sources.R
```diff
@@ -57,14 +57,25 @@ non_eu <- read_csv(here(data_dir, "raw", "non-eu.csv")) %>%
...
+            max_date = max(date)) %>%
```
sbfnk, line 69: perhaps add `.groups = "drop"` to avoid a warning

code/auto_download/hospitalisations/save-selected-sources.R
```diff
@@ -57,14 +57,25 @@ non_eu <- read_csv(here(data_dir, "raw", "non-eu.csv")) %>%
...
+  mutate(stale = ifelse(max_date < Sys.Date() - 16, TRUE, FALSE)) %>%
+  filter(stale) %>%
```
sbfnk, line 71: or just write `filter(max_date < Sys.date() - 16)`?

sbfnk (approved):
Looks good to me - made a few suggestions but I don't think they affect functionality in any way.

Re e.g. Denmark yes I think adding these to the list of exclusions for the first week after the gap would make sense.

Bisaloo (approved):

PR #1655, opened 2022-02-22
check source revisions

sbfnk:
A script to check for large revisions in our weekly data. At the moment this doesn't do anything but make plots. Once someone has reviewed if this all makes sense I think it might be good to automatically overwrite the anomalies file at the end of `check_revisions.R`.


sbfnk:
See plots inside the README at https://github.com/covid19-forecast-hub-europe/covid19-forecast-hub-europe/tree/check-sources/data-truth

sbfnk:
Forgot to say, the suggestions is to add any week to the anomalies that at any time has had a revision of greater than 5% of the largest value it has ever had.

kathsherratt:
I like this approach excluding data points with >5% change at any time - makes sense given we always (at the moment) evaluate against the latest set of data which can be revised in these small amounts but which could make a difference to performance overall.

Some thoughts (no change suggested though):

- It looks like this adds quite a few data points to our anomalies list - even excluding hospitalisations, we go from 13 to 200 exclusions. 

- Potentially, for particularly large revisions we could expand this to also exclude from evaluation the weeks after a large revision, since models could have been trained on revised data. For example of the 200 exclusions for cases and deaths, there are 43 data points that were later revised >50%. A few of these look like they have quite large numbers in absolute terms as well. But I guess at some point this gets a bit complicated and we sort of risk excluding the whole data set.

- Maybe for the paper we could use the new anomalies list for the main results, but as a sensitivity analysis also re-run evaluation with no anomalies excluded to check if this biases performance in some way.

- Not a new thought or specially relevant here, but our general approach equally discards forecasts that accurately predicted the true (later, revised) value along with those that predicted the wrong (earlier) value. It could be neat to explore if there is something useful in that information about real time forecast models that we could dig out at some point.

kathsherratt (approved):

code/auto_download/check_revisions.R
```diff
@@ -0,0 +1,125 @@
...
+      owner = owner,
+      repo = repo,
+      path = path[source],
+      date = earliest_date,
+      .limit = Inf
+    )
+
+  shas <- vapply(commits, "[[", "", "sha")
+  dates <- vapply(commits, function(x) x[["commit"]][["author"]][["date"]], "")
+  dates <- as.Date(substr(dates, 1, 10))
```
Bisaloo, line 48:
```suggestion
  dates <- as_date(ymd_hms(dates))
```

code/auto_download/check_revisions.R
```diff
@@ -0,0 +1,125 @@
...
+
+  ## keep multiples of 7 since today
+  select_commits <- which(as.integer(max(dates) - dates) %% 7 == 0)
+
+  data[[source]] <-
+    lapply(
+      select_commits,
+      function(id)
+        readr::read_csv(
+                 gsub(" ", "%20", ## for URLs
```
Bisaloo, line 58:
```suggestion
                 URLencode(
```

code/auto_download/check_revisions.R
```diff
@@ -0,0 +1,125 @@
...
+    ylab(source) + xlab("End of MMWR week")
+  ggsave(here::here("data-truth", "plots",
+                    paste0("revisions-", source, ".svg")),
+         p,  width = 20, height = 12)
+}
+
+anomalies_raw <- data %>%
+  group_by(location, location_name, target_end_date, type) %>%
+  summarise(abs_diff = max(value) - min(value),
+            rel_diff = (max(value) - min(value)) / max(value),
```
Bisaloo, line 98:
```suggestion
            rel_diff = abs_diff / max(value),
```

Bisaloo (approved):
LGTM. I've made a couple of commit suggestions. They are no functional effects but I believe they make your intent a little bit clearer.

sbfnk:
> - It looks like this adds quite a few data points to our anomalies list - even excluding hospitalisations, we go from 13 to 200 exclusions.

Yes, though still relatively small compared to the total number of data points / forecasts (~4,000, i.e. ca. 5% of data points excluded).

> - Potentially, for particularly large revisions we could expand this to also exclude from evaluation the weeks after a large revision

We're already doing this (unless I misunderstood): https://github.com/covid19-forecast-hub-europe/covid19-forecast-hub-europe/blob/4dccffec8c569dd78fc124ce52fd07f7aaa2c4fe/code/evaluation/load_and_score_models.r#L34

> - Maybe for the paper we could use the new anomalies list for the main results, but as a sensitivity analysis also re-run evaluation with no anomalies excluded to check if this biases performance in some way.

That's a nice idea (though I'm not 100% sure what we'd conclude if this yields systematic differences).

> - Not a new thought or specially relevant here, but our general approach equally discards forecasts that accurately predicted the true (later, revised) value along with those that predicted the wrong (earlier) value. It could be neat to explore if there is something useful in that information about real time forecast models that we could dig out at some point.

Again I think it's a good idea, though again I'm not 100% sure how best to investigate this.

sbfnk:
Thanks both!

PR #1666, opened 2022-02-28
Predictions 28-02-2022

BIOCOM-SC:


Bisaloo:
Hi @BIOCOM-SC :wave:, as you might have noticed, your pull requests are a bit complicated to merge because they have conflicts with our `main` branch. This happens because our GitHub Actions workflows run in your fork. Could you please disable them following [these instructions](https://github.community/t/stop-github-actions-running-on-a-fork/17965/2)?

Bisaloo:
Alternatively, you could also delete your fork and re-create it. Workflows are now disabled by default in forks.

PR #1668, opened 2022-02-28
2022-02-28-epiMOX-SUIHTER and Giovanni Ziarelli added as collaborator

giovanniziarelli:


Bisaloo (approved):
Checks fail because we don't have truth data for hospitalizations in Italy at the moment but this is not a problem to merge this submission.

PR #1678, opened 2022-03-01
2021-02-27 - JBUD submission

joebud:


Bisaloo (changes requested):
Hi @joebud :wave:, I notice you have a column `scenario_id` that contains a different value `"forecast"`. We only allow the `"forecast"` value in this column in the forecast hub. You can also delete this column entirely if you prefer.

If you want to contribute to the scenario hub, please refer to https://github.com/covid19-forecast-hub-europe/covid19-scenario-hub-europe.

PR #1684, opened 2022-03-01
Do not include exact time in reports

Bisaloo:
As to minimize conflicts if a workflow accidentally runs in a fork


kathsherratt:
Thanks Hugo, nice spot!!

PR #1688, opened 2022-03-04
Implement stricter checks

Bisaloo:


kathsherratt:
This looks good to me. I remember the early negative-value forecasts which came in before we had a check on this (although looks like some slipped through even after that). I think at the time we decided to just let them be since we didn't want to modify forecasts retrospectively. 

At this point though I'd agree with re-setting them to 0. It seems the same as when we retrospectively rounded forecasts - just rounding anything negative up to 0 - and when we retrospectively rounded forecasts, we sought everyone's permission and had no trouble with it.

sbfnk (commented):
Do we need to think about a check that the truth data is >=0, and/or add any that aren't to the anomalies? As far as I recall at the moment we can have (and have had) truth data < 0 and while it's of course nonsensical epidemiologically it feels that if we're stricter about forecasts >=0 we should do the same for truth.

Bisaloo:
I'm not sure. 

It seems to me that it's much harder to avoid negative values in truth data than in forecasts (or scenario projections). Truth data delivery is an incredibly complex system requiring continuous delivery by a system designed early in the pandemic (when all variables were not yet known) and with little flexibility and update possibility. They're bound to produce breaking changes at some point and these breaking changes can realistically translate into negative values. These negative values can be the unavoidable consequence of a system that was initially unfit / not designed ideally (even though, I believe they did the best they could with what they had at this time).
Correcting these negative values on our end seems dangerous because data providers know this is a suboptimal situation and I trust they chose to produce negative values because they assessed this is the best choice to deal with the system lack of flexibility. I don't believe we would able to make better corrections that we they did.
In short, yes, negative truth values are artifacts but unavoidable artifacts IMO. We have to accept them and work around them accordingly.

It's completely different for forecasts because they are produced at a single time point. There is no constraints regarding breaking changes in the underlying model. They is no necessity to overcorrect precedent mistakes or design choices.

sbfnk:
This all makes sense - but we should then still exclude negative values from evaluation (or perhaps we're doing so already?), as otherwise we could, in principle, be punishing forecasters by setting their predictions to 0.

Bisaloo:
It doesn't look like we're doing so at the moment if I understand this commit correctly: https://github.com/covid19-forecast-hub-europe/covid19-forecast-hub-europe/commit/9c76f48555913286d1ea5e9ddb4ef3d77a2add2b

We have negative values in Switzerland but this doesn't result in a change in the anomalies df.

Bisaloo:
Partially merged in https://github.com/covid19-forecast-hub-europe/covid19-forecast-hub-europe/commit/57fd648203cbbb35c5d6826a7d106d058743b420.

Bisaloo:
> So would a good strategy be to:
> 
>     1. Treat negative data as valid
> 
>     2. Treat negative forecasts as invalid
> 
>     3. Check for negative data in a class of checks that expands on [https://github.com/covid19-forecast-hub-europe/covid19-forecast-hub-europe/blob/main/code/auto_download/check_revisions.R](https://github.com/covid19-forecast-hub-europe/covid19-forecast-hub-europe/blob/main/code/auto_download/check_revisions.R?rgh-link-date=2022-03-07T15%3A00%3A16Z) and flags anomalous data
> 
> 
> (3) would be needed to remedy the fact that (2) is inconsistent with (1).

I'm not completely clear on this: should we flag an anomaly when there are negative values in the most recent truth data or when there are negative values at any time? What about data that is initially negative and then gets fixed? Should it be removed from the anomalies?

sbfnk:
This was somewhat forgotten but still seems like a good idea.

@kathsherratt am I right in thinking that this wouldn't create a problem with your paper on ensemble performance (except the data here would be slightly different) because you're included a data file in the github repo?

The python validation here is failing but I'm not sure why.

kathsherratt:
Thanks for looking at this Seb! 
No this shouldn't be a problem - the data are saved. The codebase has an option to load fresh data, but it's not needed at this point, and I can anyway point to a git commit history.

PR #1698, opened 2022-03-07
Collate model metadata

kathsherratt:
Collates metadata across all teams.
- Gets all the metadata in data-processed, saves as a csv
- Adds this table to the README of data-processed folder for visibility
- As teams change metadata occasionally, added a github action to refresh at 4pm Tuesdays (not an expert with gh actions, may need edits).

Suggesting this because the only way of currently accessing the full metadata across all models is to use `covidHubUtils` (or write new loopy code through data-processed) - which may not be that accessible to teams just wanting to see metadata and interact with other models. (Idea comes from updating/simplifying the model paper code.)


kathsherratt:
Thanks! I agree with all those points. The table on the website is really nice as it is. My reason for adding this in addition to the website version is to have metadata directly where forecasts are saved, and to include all the extra detail (like collaborators, contact details) which might not be obvious otherwise. Happy to add some loopy code!

kathsherratt:
This is implemented anyway in the [Hub paper code](https://github.com/covid19-forecast-hub-europe/euro-hub-ensemble/blob/main/code/load/download_metadata.R).
Whether we want to do anything more with that or package it up to make it more accessible is another question I think (@Bisaloo ) so closing this PR.

PR #1703, opened 2022-03-07
Compute baseline for 2021-12-20 instead of 2021-12-21

### LINKED ISSUE #1700: Baseline missing for forecast date 2021-12-20
Instead we have 2021-12-21-EuroCOVIDhub-baseline.csv - should this not cause a validation error?

This missing is causing some problems with some of the evaluations that are relative to the baseline. Can we re-run the 2021-12-20 version and add it to the repo?

Bisaloo:
Fix #1700


sbfnk (approved):
LGTM

sbfnk:
> LGTM

Scratch that, this looks like something went wrong?

https://forecasthub.shinyapps.io/covid19-forecast-hub-europe//?file=https://github.com/covid19-forecast-hub-europe/covid19-forecast-hub-europe/raw/88d2c79f7d246ca40f35742abfc9b9e3dc1218e7/data-processed/EuroCOVIDhub-baseline/2021-12-20-EuroCOVIDhub-baseline.csv

sbfnk:
It looks like the starting point for the baseline here is the  data point (March 2022) rather than the latest before 2021-12-20

Bisaloo:
I believe it's a bug in covidHubUtils. According to the documentation, the `truth_end_date` argument in `load_truth()` should take care of that but it doesn't.

I'll open an issue there.

Bisaloo:
If that's okay with you, I'm going to cherry-pick https://github.com/covid19-forecast-hub-europe/covid19-forecast-hub-europe/pull/1703/commits/88d2c79f7d246ca40f35742abfc9b9e3dc1218e7 in `main` so we don't end up in a strange situation where the baseline is created from two places at the same time while this PR is waiting.

sbfnk:
> If that's okay with you, I'm going to cherry-pick https://github.com/covid19-forecast-hub-europe/covid19-forecast-hub-europe/commit/88d2c79f7d246ca40f35742abfc9b9e3dc1218e7 in main so we don't end up in a strange situation where the baseline is created from two places at the same time while this PR is waiting.

Yes, that sounds good. That said, if it's not too cumbersome it would also be good to get the baseline in place sooner rather than later (e.g. by post-processing the truth data after `load_truth`) as e.g. finalising @kathsherratt's update on the forecast hub paper will have to wait for it.

sbfnk:
LGTM. So we merge the whole PR after all, right?

PR #1725, opened 2022-03-14
2022-03-14-epiMOX-SUIHTER

giovanniziarelli:


Bisaloo (approved):
Checks fail because we don't have truth data for hospitalisations in Italy but this is not a problem.

PR #1774, opened 2022-03-28
USC-SIkJalpha: Weekly update

FrostXTJ:


sbfnk:
Hi @FrostXTJ we are getting validation errors because some predictions exceeds the population size. Also some of the forecasts seem disconnected from the recent data (see preview link in the previous comment). Can you please check if something went wrong with the model?

PR #2039, opened 2022-06-14
Update plot-forecasts to scoringutils 1.0.0

Bisaloo:
I haven't implemented the limit feature yet as I'm still to figure out a good way to do it.


Bisaloo:
Merging as this is blocking #2040. The `"free_y"` option seems good enough at the moment. Even if a model forecasts a growth so steep if hides the past truth data, other models will not be affected. Overall, this seems like a good (the best?) option.

PR #2113, opened 2022-07-11
Remove panels for models without forecasts for the current date or target

Bisaloo:
Before:

image

Now:

image


Bisaloo:
Model reporting building fails in the tests because we didn't create the evaluations yet. I don't think this should block merging this branch.

PR #2155, opened 2022-07-20
Upload 2022-07-17 forecast for Karlen-pypm

deankarlen:
It has been several months since I submitted any European forecast. As a result it took longer than usual to prepare the models for this forecast. The data used to prepare these forecasts were downloaded on July 17.

The model includes three separate strains: one for each of BA.1, BA.2, and BA.4/5. The data is calibrated to hospital admission data.
Population immunity set by BA.2 turnover. Assume some escape from immunity for BA.4/5. Only 6 nations included.


Bisaloo:
Hi @deankarlen, glad to have you back on board! Since you mention you didn't use more recent data, we have agreed to make an exception and merge these forecasts even though they arrived slightly after the deadline.

PR #2157, opened 2022-07-22
Fix scientific notation values

Bisaloo:
since arrow apparently doesn't like them


Bisaloo:
This will hopefully made unnecessary by https://issues.apache.org/jira/browse/ARROW-17241 :crossed_fingers:

Bisaloo:
This doesn't seem like the correct way to fix it. Let's use `float32()` instead of `int32()` until https://issues.apache.org/jira/browse/ARROW-17241 is fixed.

PR #2238, opened 2022-08-15
Forbid addition of unnecessary scenario_id column

Bisaloo:
Code:

```r
fs::dir_ls("data-processed", type = "file", recurse = TRUE) |> 
  purrr::map(readr::read_csv, progress = FALSE, show_col_types = FALSE) |> 
  purrr::map(~ dplyr::select(.x, !tidyselect::any_of("scenario_id"))) |> 
  purrr::iwalk(~ readr::write_csv(.x, .y))
```


Bisaloo:
It's probably too late for this as we're handing the hub over to the ECDC and it would not be a nice move to let them deal with teams having to update their workflow.

PR #2254, opened 2022-08-18
2022-08-15-epiMOX-SUIHTER

giovanniziarelli:
Hi,
We had problem with our servers during this week so that we did not succeded in uploading the forecast.
Would it be possible to merge it still?
Thank you in advance,
Giovanni Ziarelli- epiMOX, Politecnico di Milano, Milan, Italy


Bisaloo:
Hi @giovanniziarelli, since you are a trusted and long-time contributor to the forecast hub, we can indeed exceptionally merge your forecasts late.

PR #2284, opened 2022-08-29
ILM-EKF forecast for 2022-08-29

stefanheyder:


RokGrah:
Hi @stefanheyder 

thank you for your submission to the COVID-19 forecasting hub. We have noticed two minor issues with your most recent submission. More specifically, forecasts for Finland and Greece show very large values for both cases and deaths. 

Would you be able to validate those forecasts to check if there are any potential issues? 

We merge all submissions into one ensemble forecast on Tuesday afternoon - it would be great we could include your forecast into it as well.

Best wishes,
Rok

PR #2382, opened 2022-10-03
2022-10-02 - SDSC_ISG-TrendModel Forecast Update

TaoSunVoyage:
Update forecast - SDSC_ISG-TrendModel


bprasse:
Hello Tao, 

The prediction of cases in CH might be off (it seems to go equal zero). Could you have a look at it? Thanks a lot!

PR #2413, opened 2022-10-10
ITWW forecast for 2022-10-10

stefanheyder:


stefanheyder:
Sorry for the many updates - migrating from the old DE/PL ForecastHub so this submission took a while. Hope everything works as expected now!

PR #2441, opened 2022-10-17
Update 2022-10-17-AMM-EpiInvert.csv

lalvarezmat:


lalvarezmat:
Hello,

We have corrected a small mistake in the Germany forecast in 
2022-10-17-AMM-EpiInvert.csv. So I updated the file 
2022-10-17-AMM-EpiInvert.csv. I thought that we could modify the current 
forecast as far as we do it on Monday isn't it?

Regards
Luis

Issue #2446, opened 2022-10-18
Incidence of Hospitals in Slovenia are wrong 


janezz25:
Dear All,

We are observing some huge discrepancies on number of hospital admissions for Slovenia. The numbers should be at least a magnitude lower than these, which are shown in forecasts for Slovenia. 

Please see https://covid-19.sledilnik.org/en/stats under the plot Hospitalizations.

sbfnk:
Thank you! This should be fixed now.

PR #2451, opened 2022-10-22
hospitalisation source update

sbfnk:
Updating as scraped data no longer updated. Further revisions likely to come but this improves things for now.


sbfnk:
Fixes #2446

PR #2486, opened 2022-11-01
Adds forecast ITWM

neele-itwm:


RokGrah:
Hi @neele-itwm, thanks for your submission to the forecasting hub. After checking it, I see the values for cases and deaths are forecasted to be much lower than in the past weeks. Would you be able to check if all is ok on your end?
Many thanks,
Rok

PR #2525, opened 2022-11-14
ITWW forecast for 2022-11-14

stefanheyder:


RokGrah:
Hi @stefanheyder, thanks for your submissions. I see that the forecasted death count for Germany is much lower than what was seen in the past. Would it be possible to check if all is ok with the model and data? 
Many thanks,
Rok

PR #2675, opened 2023-01-03
2023-01-02-HZI-AgeExtendedSEIR

istirodiah:


kathsherratt:
Hi @istirodiah, thank you very much for the forecast. (Unfortunately as it's a bit late it won't be included in the ensemble this week.)
Just to note, in future submissions it would be great if you could remove the `scenario_id` column, as this isn't really necessary here.
Thanks again,
Kath

PR #2699, opened 2023-01-10
Set actions to only run from this organisation

kathsherratt:
We [continue](https://github.com/covid19-forecast-hub-europe/covid19-forecast-hub-europe/pull/2681) to have problems with Actions running on forks of the repo. These are difficult to control because it requires the fork owner to disable the actions. 

I suggest we add an if statement to all the Actions so they only run when our org (covid19-forecast-hub-europe) is the repo owner. (Following other people's experience of this [here](https://github.com/orgs/community/discussions/26704?sort=top))

Adding one example here then can copy to other actions if it seems like a good idea.

### DIFF
--- a/.github/workflows/ECDC.yml
+++ b/.github/workflows/ECDC.yml
@@ -7,6 +7,7 @@ on:
 
 jobs:
   get_ecdc:
+    if: github.repository == 'covid19-forecast-hub-europe'
     runs-on: ubuntu-20.04
     env:
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/JHU.yml
+++ b/.github/workflows/JHU.yml
@@ -6,6 +6,7 @@ on:
 
 jobs:
   get_jhu:
+    if: github.repository == 'covid19-forecast-hub-europe'
     runs-on: ubuntu-20.04
     steps:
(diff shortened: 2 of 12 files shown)


sbfnk:
This does seem like a good idea! I think in the past @Bisaloo argued that this wasn't needed and therefore [removed the checks](https://github.com/covid19-forecast-hub-europe/covid19-forecast-hub-europe/commit/dd71de1a893a86c795fcd6ed29f6f3141b9834c8) but your example seems to indicate that this is wrong so I'd suggest that you go ahead and bring the checks back.

Bisaloo:
I confirm that this is not needed with the current GHA framework. Issues that you seeing are due to forks that were created before GitHub made the change of automatically disabling workflows on forks.

The proper fix IMO would be to ask contributors to delete and re-create their fork. I've done this a couple of times in the past when I was merging submissions and it did solve the issue.

I understand if you prefer having this workaround instead of asking extra work from contributors though. Just note that it comes with its own set of issues, for example when you rename the organisation (which is likely to happen at some point, right?)

kathsherratt:
Thanks for flagging Hugo, I agree that is really the more proper fix. I think a few contributors are pretty much running automated weekly submissions themselves though so comments get missed unfortunately and I'm keen to just nip this in the bud at this point.

Issue #2634, opened 2022-12-19
reports-model workflow status


Bisaloo:

kathsherratt:
Adding here that we might want to wrap the call to [render model reports](https://github.com/covid19-forecast-hub-europe/covid19-forecast-hub-europe/blob/main/code/reports/compile-model-reports.r) in a `safely()` function, so that reports are still created even if one model's reports fail the other model reports are still created (plus an informative error noting which models failed / where / why)

sbfnk:
> Adding here that we might want to wrap the call to [render model reports](https://github.com/covid19-forecast-hub-europe/covid19-forecast-hub-europe/blob/main/code/reports/compile-model-reports.r?rgh-link-date=2023-01-19T12%3A54%3A48Z) in a safely() function, so that reports are still created even if one model's reports fail the other model reports are still created (plus an informative error noting which models failed / where / why)

Good idea!

sbfnk:
fixed in https://github.com/covid19-forecast-hub-europe/EuroForecastHub/commit/eb009198c9b4f6a76f44beb5181408398a5914a4

PR #2721, opened 2023-01-17
add snapshots of historic OWID hospitalisation data

sbfnk:
Adds historical OWID hospitalisation data by running

```r
source(here::here("code/auto_download/hospitalisations/get-owid-historical.r"))
get_owid_historical(earliest_date = lubridate::today() - months(6))
```

Also adds revision checking and a recommended cutoff based on the number of recent weeks that are usually subject to >5% revision, and updates plotting routines to highlight data likely subject to revisions.


kathsherratt (approved):
This looks great, thank you Seb :)

kathsherratt:
Noting that this seems to introduce a new issue with the model reports - [failing](https://github.com/covid19-forecast-hub-europe/covid19-forecast-hub-europe/actions/runs/3958275598/jobs/6779647294) at the `forecast-calibration.Rmd` chunk with the error "After removing all NA true values and predictions, there were no observations left".

(not sure if it applies here, but if the error is just for one model then issue #2634 might help debug this)

PR #2723, opened 2023-01-18
Hospitalisation data download - use NA values to replace empty df

kathsherratt:
Updates `save-selected-sources.R` to add a single row of NA values if hospitalisations dataframe is empty.

I was thinking that NA values can be operated on unlike an empty df, and this might (possibly) help fix some of the workflows dependent on some data being present in the hospitalisations truth data (which currently returns 0-row dataframe when all data are filtered out).

Saying that though, the hosp data has just updated successfully [with data](https://github.com/covid19-forecast-hub-europe/covid19-forecast-hub-europe/blob/main/data-truth/ECDC/truth_ECDC-Incident%20Hospitalizations.csv), and hopefully we will be switching data sources soon anyway.


sbfnk (commented):
This make sense but I worry a bit that this could cause problems in other parts of the workflow - and if there is no data we probably want to know. Would it perhaps make more sense to throw an informative error here?

sbfnk:
Also tried to [fix at the source](https://github.com/reichlab/covidHubUtils/pull/388) but hopefully this won't happen again.

PR #2754, opened 2023-01-30
2023-01-29 - SDSC_ISG-TrendModel Forecast Update

TaoSunVoyage:
Update forecast - SDSC_ISG-TrendModel


kathsherratt:
The fork needs to be brought up to date with main, e.g. re-cloned, to avoid multiple file creation from github actions (this problem is fixed on main). I don't have write access to the fork so can't complete this.
I have taken the forecast file in this PR and submitted it in #2773

PR #2869, opened 2023-02-27
Fix concurrency limitation in submission_preview.yml

Bisaloo:
I am going through the docs for this feature again for another project and I think I figured out why pushes across different PR could sometimes cancel the workflow running in another PR.

Since `submission_preview.yml` is triggered by `pull_request_target`, `${{ github.ref }}` will always be `main`.


sbfnk:
Thanks!

PR #2902, opened 2023-03-07
2023-03-06 fjordhest-ensemble

fhi-kan:
New team, first submission


kathsherratt:
Hi @fhi-kan and team, 

First of all thanks so much for being our newest contributor to the Hub! We really appreciate the effort it takes to create and submit forecasts and it will be great to include yours.

Thanks also for following the submission instructions. At the moment there are a few errors in the validation process. I'm sorry for the unclear error messages produced earlier - the new error messages produced by the `Validations (R)` check are more informative (copied below).

I think you will need to do the following to pass the validations:
- In the metadata file:
   - `website_url` - please fill this in with a text url, e.g. a github repo from where you submit the forecasts or a link to your organisation's website. 
   - `repo_url` - this is optional, if you don't use then just remove it. 
   - `team_model_designation` field - this should be all lowercase 
- In the forecast file:
   - `value` - as we're forecasting counts, we ask modellers to submit only integer values.

Let me know if I can help any further and looking forward to including your forecasts.

Kath



---

```
The validation checks produced some errors reported above.
 - /website_url must be string
Execution halted
 - /team_model_designation must be equal to one of the allowed values
 - /repo_url must be string
! 2023-03-06-fjordhest-ensemble.csv: Data has to be formed of the expected columns with correct type 
 - /value/0 must be integer
```

fhi-kan:
Hi Kath,

Thanks for the welcome and for coordinating this important effort.

We have updated the metadata file (removed repo_url, added website_url, changed case of model_designation) and rounded 'value' in the forecast file. Additional issues appear to exist.

Thanks,
Sasi.

kathsherratt:
Hi Sasi, 
Thank you for making those changes. I'm not sure why one validation check (run in Python) fails when the other check (in principle the same but run in R) now passes. It looks like it is picking up something in the metadata but all looks fine to me. I will merge this in any case as the forecast is valid.
Kath

fhi-kan:
Thank you!

PR #2974, opened 2023-04-09
SGroup-RandomForest Submission for April 9th

maa989:


kathsherratt:
Hi Majd @maa989 ,

Thanks so much for contributing a new forecast! 

The failing check is an error on our side with metadata validation - you don't need to take further action.

Kath

PR #3012, opened 2023-04-24
2023-04-24 fjordhest-ensemble

fhi-kan:
Truth data for cases and deaths does not seem to have been updated since last week.


sbfnk:
Thanks for flagging, @fhi-kan. I've [raised this](https://github.com/EU-ECDC/COVID-19_weekly-data/issues/3) with the data providers.

PR #3057, opened 2023-05-22
retrospective baseline update

sbfnk:
Due to an error in the baseline model (thanks to @RokGrah for spotting it!) when it was updated to the changed data sources it kept making forecasts off the final JHU data. This was fixed in https://github.com/covid19-forecast-hub-europe/covid19-forecast-hub-europe-baseline/commit/2721323f54b17e06e9c12e45cfac19855ddff5cb

It's generally not great to update past models but since the baselines are used for model evaluation I think it makes sense to retrospectively change them. Since the baseline model is now past of the ensemble, it also updates past ensembles. This PR makes these changes.


kathsherratt:
Thank you so much for finding and fixing this! I agree it is necessary to update the past baseline forecasts. Strictly speaking I guess we don't need to update the ensemble (I think in the past we haven't when we found bugs) but I agree it's alright to for this. Thank you!!

Issue #2636, opened 2022-12-19
visualisation workflow status


Bisaloo:

Bisaloo:
https://www.githubstatus.com/

>  Update - We are investigating reports of issues with service(s): multiple services. We will continue to keep users updated on progress towards mitigation.
Jan 25, 2023 - 08:30 UTC
Update - Actions is experiencing degraded performance. We are still investigating and will provide an update when we have one.
Jan 25, 2023 - 08:04 UTC

Issue #2630, opened 2022-12-19
check-truth workflow status


Bisaloo:

kathsherratt:
Fixed in #2676

sbfnk:
Fixed in 655501b137abeccb34ffd3be80894034730b9186

Issue #2628, opened 2022-12-19
ECDC workflow status


Bisaloo:

kathsherratt:
Fixed in #2678

sbfnk:
Fixed in https://github.com/covid19-forecast-hub-europe/covid19-forecast-hub-europe/commit/0fe28d1e56dbf1f370fecb9207ff3838a84a542c

sbfnk:
Fixed in #3365

PR #3137, opened 2023-07-12
update renv to 1.0.0

sbfnk:
Some Github actions are currently failing. It seems that this is happening because they are downloading renv 1.0.0.

This PR is to test if this can be fixed by upgrading renv.


sbfnk:
Closes #2630 
Closes #2628 
Closes #2634 
Closes #2636 
Closes #2631

PR #3319, opened 2023-12-08
Update death data source

sbfnk:
This updates the source of death data to the new github repo for respiratory virus data at
https://github.com/EU-ECDC/Respiratory_viruses_weekly_data/tree/main

It also deprecates case data as these are no longer available.


sbfnk:
Update: turns out case data are available but in a different file

PR #3376, opened 2024-01-15
Submission week 3

reneniehus:


reneniehus:
@sbfnk , I am wondering if you could have a look at the above failed validation run (https://github.com/european-modelling-hubs/covid19-forecast-hub-europe/actions/runs/7533334153/job/20505654866?pr=3376). I cannot identify the issue. Many thanks

sbfnk:
I think it's not an issue with the model but some issue with R on the Github Action node. Just tried re-running the validation action in the first instance.

PR #3396, opened 2024-01-29
2024-01-29-PL_GRedlarski-DistrictsSum

gr3gr3:


reneniehus:
Hi @gr3gr3 , many thanks for your pull request. Your pull request contains commits from the past 4 weeks. This means that your old forecasts could not be part of the hub ensemble - just so you are aware. These seems to have been an issue so that in previous weeks you closed your own pull requests before they could be merged. What was the particular reason? Is there anything we can support you with?

